PR #75, opened 2014-09-26
Added type check for compatibility with datetime.datetime

treilly:
Added type check so assert isn't triggered by a datetime.datetime without a microsecond property


fzumstein:
OK so after all it did make sense but that way it was only working on 2.7 -- implemented a slightly different fix for v0.3.3

PR #105, opened 2014-11-03
Remove sheet and some Range methods

kvdogan:


fzumstein:
Remove sheet replaced by #297 - rest already implemented.

PR #118, opened 2014-11-27
added 'lastrowandcolumn'

dutch-rob:
Added a function lastrowcol in main.py to give the row number and column
number after which there is no more content on the worksheet, and the
corresponding implementation in win32com/windows and appscript/osx.
We discussed this a bit on stackoverflow. I started with working from
'xlLastCell' in Windows because that would prevent looking at rows or
columns with only deleted content, but I could not figure out how to do
that in appscript, so I used 'usedrange' there. I read somewhere that
usedrange also includes rows/columns with deleted content though when I
tried, that did not appear to be the case.

### DIFF
--- a/xlwings/_xlmac.py
+++ b/xlwings/_xlmac.py
@@ -69,6 +69,13 @@ def get_worksheet_name(xl_sheet):
     return xl_sheet.name.get()
 
 
+def get_worksheet_last_row_and_column(xl_sheet):
+    ur = xl_sheet.used_range
+    nrow = ur.first_row_index.get() + ur.count(each=kw.row) - 1
+    ncol = ur.first_column_index.get() + ur.count(each=kw.column) - 1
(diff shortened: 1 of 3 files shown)


fzumstein:
What about if we call that `Sheet.last_cell()`, have it returns a `Range` object and implement the `Range.row` and `Range.column` properties analogous to Excel VBA so you can get the row and column number of that last cell?

fzumstein:
or maybe one step more: have `Sheet.used_range` return the used range as `Range`, then add the `last_cell` property to `Range` along with `column` and `row` so you could do:

    >>> last_cell = Sheet.used_range.last_cell
    >>> last_cell.row, last_cell.column
    (23, 34)

dutch-rob:
Hi Felix, I apologize for the delay, but your idea seems great to me!
Rob

PR #151, opened 2015-02-12
Use Tox for testing on Windows as well

### LINKED ISSUE #150: Tox fails to run on Windows due to tox.ini deps
As the subject says. Can't compile appscript on Windows due to missing Carbon headers.

TFenby:
tox.ini references requirements.txt, which references setup.py, which
resolves the appropriate install_requires depending on sys.platform.
Closes #150


TFenby:
Put a pause on this. Python 3.4 is giving issues with numpy because of differences in VC++ compilers. Working to figure it out onw.

fzumstein:
Ah right so that's the deal - you have to compile it from source to make it work with pip/tox, correct?

TFenby:
Well, part of it is C, and they don't provide wheels for Windows for some reason (I'm looking into that too), so you do need a compiler. MS has one for Python 2.7 packaged and ready to go ([here](http://www.microsoft.com/en-us/download/details.aspx?id=44266)), but none for Py3k that I've found yet.

fzumstein:
I would love to get this somewhow working with `conda` - so that numpy is installed via conda, not via pip.

TFenby:
Hmm. https://tox.readthedocs.org/en/latest/config.html#confval-install_command=ARGV

fzumstein:
Will look into it. This might also help: http://stackoverflow.com/questions/24832693/error-when-using-source-in-a-python-tox-ini-file

fzumstein:
I looks like you haven't set your email correctly with git. I merged your other PR and you show up in the commits, but your name is not linked to your GitHub profile and therefore you also don't show up under contributors...

TFenby:
Eh, I'm not overly concerned about it if you aren't.

fzumstein:
Sure it's cool - it's just unfortunate that you don't get "credited" and GitHub doesn't include you in the Contributor count.

fzumstein:
I guess the new platform option in tox 2.0 helps with these issues here? https://tox.readthedocs.org/en/latest/config.html#confval-platform=REGEX

TFenby:
That makes the platform selection more straightforward, but doesn't solve the issue w/ some wheels not being available on Windows.

fzumstein:
Right - I think this might be solved by downloading the binaries for Windows and having them install via easy_install (by setting the install_command: http://tox.readthedocs.org/en/latest/config.html#confval-install_command=ARGV)

fzumstein:
I am closing this as I have replaced this with #377 - I moved away from tox/pip to conda envs. It's cross-platform, much faster to setup the envs, no need to install all interpreters and no issues with any of the dependencies (especially the new ones like matplotlib and pillow)

PR #152, opened 2015-02-12
Headless/invisible window support

### LINKED ISSUE #149: Any issues with adding visible parameter to Workbook init?
I'm looking to use this for processing data in the background, and opening a workbook then immediately setting it as not visible is ugly. Any issue with adding visible=True to Workbook's init parameters, then passing that on to the various *_workbook functions?

TFenby:
Adds the visible parameter to Workbook's `__init__`, defaulting to True to match the previous behavior. False makes the application invisible, and None does not alter its visibility.

Closes #149 

Please note that this is a bit of a hack, as it isn't a Workbook that's visible or not, it's the entire Excel app. However, because Workbook is currently the highest-level object, it was the most logical place to put this functionality.

OS X support needs to be implemented still, as I do not own a system to test it on.

### DIFF
--- a/xlwings/_xlmac.py
+++ b/xlwings/_xlmac.py
@@ -389,3 +389,7 @@ def save_workbook(xl_workbook, path):
 
 def open_template(fullpath):
     subprocess.call(['open', fullpath])
+
+def set_visible(xl_app, visible):
+    # FIXME: Implement
+    pass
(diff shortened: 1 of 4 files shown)


TFenby:
This doesn't depend on this Tox PR, though, and is still g2g.

PR #164, opened 2015-03-17
Added exception when referencing (0,0); resolves #106

### LINKED ISSUE #106: Throw exception when Range(0,0) called
This is probably a common mistake since Excel is 1 based on Python 0 based. Currently there isn't an explicit Exception thrown...

ntdef:
Throws an exception when a user tries to reference cell as if it was base 0, in response to bug #106.

### DIFF
--- a/xlwings/main.py
+++ b/xlwings/main.py
@@ -675,6 +675,8 @@ class Range(object):
             self.row2 = self.row1 + xlplatform.count_rows(self.xl_sheet, range_address) - 1
             self.col2 = self.col1 + xlplatform.count_columns(self.xl_sheet, range_address) - 1
 
+        if 0 in (self.row1, self.col1, self.row2, self.col2):
+            raise IndexError("Attempted to access cell at index-0. Excel ranges are all all base-1 index")
         self.xl_range = xlplatform.get_range_from_indices(self.xl_sheet, self.row1, self.col1, self.row2, self.col2)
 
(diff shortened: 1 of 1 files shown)


xlwings/_xlwindows.py
```diff
@@ -135,6 +135,8 @@ def count_columns(xl_sheet, range_address):
 
 
 def get_range_from_indices(xl_sheet, first_row, first_column, last_row, last_column):
+    if 0 in (first_row, first_column, last_row, last_column):
```
fzumstein: This check should be done in `main.py` as otherwise you need to duplicate things for Mac.

fzumstein:
Hi Troy, did you want to fix this PR (see comment from back then)? Otherwise I ll close it. Thx

ntdef:
@fzumstein Yup, sorry to leave you hanging: I'll fix this asap.

ntdef:
@fzumstein Ok I've updated my PR.

fzumstein:
There must have been something gone wrong: your PR deletes _xlwindows.py and xlwings_tests.py

ntdef:
Fixed. Try it now: I removed some of my older changes in those files and for some reason git dropped them.

fzumstein:
merged now - I don't see the necessity of the second check  - also added tests and reworded slightly: 46035146c0f64ab992277d74ccb87a9430fa2339

PR #238, opened 2015-09-30
Embedding AddComment function

alekz112:


fzumstein:
Thanks for that. However, it should be extended so that comments can also be read back again. Excel VBA has a bit of a weird object model there so we might want to think about how to implement that in xlwings.

alekz112:
I just don't need to read comments at the moment, so I did not think about it. Good point though, I added that feature, so now you can read comment as well by Range("someRange").comment().

fzumstein:
I guess I would implement it as property, but thing is, we then also need to add a possibility to delete it again ;) it's similar to the Range.name functionality...

alekz112:
I see what you mean, I've rewritten it a bit, please, have a look :)

fzumstein:
I think we'll need to be closer to the original object model (https://msdn.microsoft.com/en-us/library/office/ff196533.aspx), i.e. make the property return an obect.

fzumstein:
Also, can you please use more specific commit messages than just "Update file x". thanks.

fzumstein:
this is pre 0.9 refactor so has to be rewritten - see #896

Issue #258, opened 2015-10-17
bug if PYTHONPATH contains multiple directories


sdementen:
If PYTHONPATH contains multiple directories separated by ";" the RunCommand is not correct.

        RunCommand = "python -c ""import sys; sys.path.append(r'" & PYTHONPATH & "'); " & Command & """ "

can be replaced by

        RunCommand = "python -c ""import sys; sys.path.extend(r'" & PYTHONPATH & ".split(";")'); " & Command & """ "

sdementen:
Hello Felix, I did a PR for both issues 258 and 257. I am not familiar yet with PRs, did I do that correctly ? Do I need to do an extra step ?

fzumstein:
Regarding the PRs, there's nothing really wrong, but I still have the following inputs:

- Name PRs according to what they to, i.e `Added app property` rather than `proposal for issue #253`
- Same goes for the underlying branch, better to name it `app-property` than something like `patch1` - but I think you've already done that in the more recent PRs
- If possible and if it makes sense, add a unit test

fzumstein:
Just noticed now that this has to be fixed for Mac and OptimizedConnection also

fzumstein:
Fixed by 93cd8f6d830ac0d047b86d828d91f9d5c8566ecd

xkeecs:
did a google search and found this solution. so in the new version we can separate multiple python path by ";". I think it's better to change the docs: http://docs.xlwings.org/en/stable/addin.html here to make the description of PYTHONPATH consistent with UDF_MODULES.

Issue #257, opened 2015-10-17
issue with PYTHON_WIN with blank space in windows


sdementen:
If PYTHON_WIN has blanks in it, the cd command fails as there is a need to quote PYTHON_WIN

PR #259, opened 2015-10-17
VBA module bugfix/enhancements (issues #258 and  #257)

sdementen:
Add quotes around PYTHON_WIN (necessary if blanks/spaces in path)
Change sys.path adaptation to handle multiple folder in PYTHONPATH.

If OK, it remains to update all xlsm files (template+examples). Is there a tool that does that automatically ?


fzumstein:
Nope, no tool yet for updating the VBA module...

Issue #254, opened 2015-10-16
add a context manager to easily "freeze" excel when running code


sdementen:
It is often useful to freeze excel while running some code, freeze in the sense of disabling (=setting to False):
- Application.ScreenUpdating
- Application.EnableEvents
- Application.DisplayAlerts
- setting calculation mode to manual

A nice API would be a simple context_manager in the vein of:

```
with app.freeze(events=True, screen=True, alerts=True, calculation=True):
    ... do some expensive changes in a workbook ...
```

that would save the status of these variables, change them according to the flags, yield and then set back the variables back to their original status.

Any feedback on the API ? if ok, i'll do a PR for the Windows version (I have no experience nor access to a Mac)

fzumstein:
This seems like a good idea, and the API can still be discussed/fine tuned before the first release...

zogzog:
app.interact could be slightly less confusing than app.freeze but I like the idea.

sdementen:
What about 
```
with app.enable(events=False, screen=True, alerts=False, calculation=False):
    ... do some expensive changes in a workbook ...
    ... events, alerts and calculation are disable/frozen
```
We could also add the Application.Interactive attribute as app.interactive to come to for example:
```
with app.enable(calculation=True, interactive=False):
    ... do some expensive changes in a workbook ...
```

fzumstein:
implemented in #1671

PR #260, opened 2015-10-17
add a context manager to easily "freeze" excel when running code (issue #254)

sdementen:
still need to update documentation about release version (is this automated ?)

### DIFF
--- a/docs/whatsnew.rst
+++ b/docs/whatsnew.rst
@@ -30,7 +30,7 @@ Also, there are some new docs:
 Bug Fixes
 *********
 
-* The Excel template was updated to the latest VBA code (:isse:`234`).
+* The Excel template was updated to the latest VBA code (:issue:`234`).
 * Various issues with timezone-aware objects were fixed (:issue:`195`).
 * [Mac Version]: A certain range of integers were not written to Excel (:issue:`227`).
 
--- a/docs/whatsnew.rst
+++ b/docs/whatsnew.rst
@@ -31,6 +31,7 @@ Bug Fixes
 *********
 
 * The Excel template was updated to the latest VBA code (:issue:`234`).
+* Connections to files that are saved on OneDrive/SharePoint are now working correctly (:issue:`215`).
 * Various issues with timezone-aware objects were fixed (:issue:`195`).
 * [Mac Version]: A certain range of integers were not written to Excel (:issue:`227`).
 
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     description='Make Excel fly: Interact with Excel from Python and vice versa.',
     long_description=readme,
     data_files=data_files,
-    packages=['xlwings', 'xlwings.tests'],
+    packages=['xlwings', 'xlwings.ext', 'xlwings.tests'],
     package_data={'xlwings': ['*.bas', 'tests/*.xlsx', 'xlwings_template.xltm']},
     keywords=['xls', 'excel', 'spreadsheet', 'workbook', 'vba', 'macro'],
     install_requires=install_requires,
--- a/xlwings/__init__.py
+++ b/xlwings/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 import sys
 
-__version__ = '0.4.1'
+__version__ = '0.4.2dev'
 
 # Python 2 vs 3
 PY3 = sys.version_info[0] == 3
--- a/xlwings/_xlmac.py
+++ b/xlwings/_xlmac.py
@@ -489,6 +489,22 @@ def get_screen_updating(xl_app):
 def set_screen_updating(xl_app, value):
     xl_app.screen_updating.set(value)
 
+def get_enable_events(xl_app):
+    return xl_app.enable_events.get()
+
+
+def set_enable_events(xl_app, value):
+    xl_app.enable_events.set(value)
+
+
+def get_display_alerts(xl_app):
+    return xl_app.display_alerts.get()
+
+
+def set_display_alerts(xl_app, value):
+    l_app.display_alerts.set(value)
+
+
 
 # TODO: Hack for Excel 2016, to be refactored
 calculation = {kw.calculation_automatic: Calculation.xlCalculationAutomatic,
--- a/xlwings/_xlwindows.py
+++ b/xlwings/_xlwindows.py
@@ -79,8 +79,11 @@ def get_excel_hwnds():
 
     excel_hwnds = []
     for hwnd in hwnds:
-        if win32gui.FindWindowEx(hwnd, 0, 'XLDESK', None):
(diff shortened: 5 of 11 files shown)


fzumstein:
Nothing automated here neither, but it's OK the way you have it, I ll adjust to the working version whenever I ll merge.

fzumstein:
I would like to keep the kwargs of `freeze` closer to Excel's version: 

`calculation_manual` instead of `calculation`
`enable_events` instead of `events`
`screen_updating` instead of `screen`
`display_alerts` instead of `alerts`

I think it makes it much easier to adopt when coming from VBA. In any case, currently `calculation=True` is highly confusing: I would expect it to actually *do* calculations when I see that statement.

sdementen:
As the function is called "freeze", it make more sense to me to write "freeze(calculation=True)" to mean "do freeze the calculation". And in this spirit, the name "events" instead of "enable_events" is also logic (meaning "freeze events"). Besides, the excel naming is not that consistent (why do not we have "enable_alerts" instead of "display_alerts" ?).

If you want to have the original names, the context manager should not be called "freeze" but something like "set_behavior" and then you can repeat all the names as in excel ("enable_events", etc). I would then avoid a calculation_manual as it is not consistent with excel (that uses simply calculation).

But I think that, in general, the "freeze" functionality is the one needed, ie, we want normally to have Excel functioning as usual and then, temporarily (in a context manager), freeze some of this dynamic behavior to do some heavy operations.

fzumstein:
Ok fair enough. However, Eric and I were thinking about moving such non-core functionality into something like an `ext` subpackage: `from xlwings.ext import freeze`

sdementen:
Could be an option indeed but then it would not be a method of Application
(no monkey patching) so we could not write

with app.freeze(events=True):

but we would have to write

with freeze(app, events=True):

which IMO looks less readable (but not that a big deal)


On Wednesday, December 2, 2015, Felix Zumstein <notifications@github.com>
wrote:
> Ok fair enough. However, Eric and I were thinking about moving such
non-core functionality into something like an ext subpackage: from
xlwings.ext import freeze
>
> —
> Reply to this email directly or view it on GitHub.<
https://ci6.googleusercontent.com/proxy/NB9radvuYpnFdL2n167aAtbOSRaMQRhbZRY3wEB6uWKeQUP6bhJjwQr9CjEjAReqrhr1rdqm7uA3mXuHR80DMMLwnygKZKCpXtKZ_rEHKTCkvBJ8DRdW-bwc5F7Us3QiYaULNmXa30ws-P9WMypfZyPIUVFMXQ=s0-d-e1-ft#https://github.com/notifications/beacon/ABPpdt2OBvveub74_4CKGc-5tDDhIb77ks5pLr4jgaJpZM4GQrhI.gif
>

ericremoreynolds:
Hey Sebastien,

why would we necessarily exclude monkey patching?

I don't see that it would create so many problems - to use `app.freeze` u just need to make sure that you have done `import xlwings.ext.freeze` at some point.

sdementen:
hello Eric,

I have no fundamental problem with monkey patching when related to obscure
behind the back behavior to tweak.

But I wonder if for user functionalities, this wouldn't be surprising. If
he/she sees a "with app.freeze(events=True):" in some code on stackoverflow
and tries it at home, he will be welcomed by a "attribute does not exist"
and it would take (i think) some time to see that it is a side effect of an
import statement.

On the hand, if we want to provide a way to "extend" the core xlwings
functionalities but in some integrated way (avoiding for instance code like
"with freeze(app, events=True):"), it may be the only way to do it.

Now, the "freeze" functionality is more a "core" one as I think (but I may
be wrong) it is the only "use case" for tweaking the 4 behavioral functions
(enable_events, displayalerts, ...). At least, in my use, I always change
these because I want them to be "disabled/frozen" during some operation and
then be "enabled/unfrozen" directly after (otherwise, the behavior of Excel
starts to be quite weird).

ericremoreynolds:
Ok let's go for `with freeze(app, events=True)` then.

Sorry to insist on the idea of keeping it a separate in an `xlwings.ext` module, the idea is that since we're still actively developing xlwings, we'd rather not have this within the perimeter of code we're directly responsible for. Keeping it in an `ext` module marks it as being contributed code.

In fact, if it's ok, in practice you can maintain this code. Please put your name and details in a comment at the top of the source file (something like [this](https://github.com/git/git/blob/master/contrib/subtree/git-subtree.sh)) and then should changes to the xlwings API break the code, you can fix it via a PR.

And of course we can repeat this mechanism for any other features you'd like to contribute - this should help us accept them more quickly.

How does that sound?

sdementen:
OK, let's have a try !

But I could not follow exactly what was meant by the document on subtrees
you linked so I am a bit lost on what to do.

Would it work like this:
1) I create a repo on my github named sdementen/xlwings.ext.UIHelper (and
so now I need to find a name for such extension !)
2) you link it from the original xlwings repo
is that more or less what happens ? (I have found this
https://help.github.com/articles/about-git-subtree-merges/ and read it
quickly)

What happens with testing, the release to PyPI and the documentation ? is
it more or less integrated with xlwings or totally separated ? if
separated, it may be a bit overkill for freeze ;-) but adapted for more
heavy-weighted extensions.

Do you know if it is that way that sqlalchemy.ext or SDL2.ext works ?

fzumstein:
Hey Sebastien,

Eric was only referring to the file structure, not the content of subtree. You don't have to change anything in how you contribute via PR. Just move things into `ext` and put your name at the top of the module.

sdementen:
ok, got it ! I'll do that.

ericremoreynolds:
Yes, sorry! The subtree link was just an example of putting your name at
the top of the file in a comment.

ericremoreynolds:
Just to show that contributed code is managed this way in other open source
projects too...

sdementen:
Eric, could you have a look at the new commit ? does it go in the right direction ?

ericremoreynolds:
Hi Sebastien,

Yes that's definitely the sort of thing I had in mind.

Thanks for all the hard work, as soon as have time we'll merge your PR.

Eric

sdementen:
Do you plan for ext code to be imported explicitly by the user to get access to it ? Or would you import the ext code directly (and there monkey patching would make sense) to make it transparent to the user (but nevertheless keep the code base separated)?
I was thinking the first scenario when writing previous post. We're you thinking the second one ?

wkschwartz:
This looks like a nice feature that hasn't received any attention in a couple years. What next steps can I help with to get this merged?

sdementen:
You can look at issue #254 as there were some new comments re the API. 
And there is the discussion on the fact it should be in an extension package. I feel to be a bit overkill but it is up to Felix/Eric to decide

fzumstein:
replaced by #1658

Issue #253, opened 2015-10-16
access application object from Sheet or Range


sdementen:
It could be useful to get access to the application object (from xlwings not the xl_app) from a Sheet or a Range object (something like sh.app or sh.application or sh.parent).

(Felix, i will post a couple of small API "enhancement" through github, is it ok for you ? or do you want me to use UserVoice ?)

sdementen:
btw, I can do PR to implement them but I'd like to have opinion/advice before doing so

fzumstein:
Sure, sounds good! And posting them here makes sense, too.

sdementen:
is it possible for me to make an issue as "enhancement" ? or is it only the "owner" of the git repo that can do that ?

fzumstein:
Yeah only the owners can add the `enhancement` label

Issue #236, opened 2015-09-27
OPTIMIZED_CONNECTION=True may lead to ghost processes


sdementen:
On 0.4.1 with OPTIMIZED_CONNECTION = True, I call the buggy_function as defined herebelow

    from xlwings import Workbook
    
    app = None
    
    def buggy_functions():
        global app
    
        """ produces standard normally distributed random numbers with shape (n,n)"""
        wb = Workbook.caller()  # Creates a reference to the calling Excel file
        app = wb.xl_app
        #app = None

When I quit Excel, both Excel and the pythonw process stay alive. I must kill Excel in the Windows Task Manager to kill both.

If the last line is uncommented (cancel the global reference to the app), processes are properly killed when exiting Excel.

fzumstein:
Welcome to COM hell. Same effect should happen when you place `wb = Workbook.caller()` outside of the function. For some reason it doesn't manage to garbage collect all the Workbook references correctly and therefore doesn't end the Python process. We haven't found the real reason yet. However, the current workaround of not using a module-wide global with the workbook or app reference seems to work reasonably well.

sdementen:
Indeed, clearing the référence 
Would it be possible to kill the python process before/when closing excel ?
Or check every second from python if app.Visible=true and if not kill python and excel afterwards ?

fzumstein:
we also allow for `app_visible=False`, so that check wouldn't work. Our plans are actually to integrate ExcelPython fully in the near future, so then we'll give it another look for sure. Until then, things stay experimental ;-)

sdementen:
When you say integrate with ExcelPython, what do you have in mind? Use the connection/interface with excel from ExcelPython and use the excel API from xlwings ? If you have a link to some URL talking about this (email thread, wiki page, ...), I'd be interested !

fzumstein:
No formal issue opened right now, but it's mostly about adding support for UDFs - and in fact most of it is already there. What's missing is the developer add-in to be able to import UDFs that are decorated in Python. A few edges need to be ironed out like those you've already spotted and some other things need to be refactored so that e.g. we can use the pandas DataFrame handling for UDFs, too.

PR #261, opened 2015-10-17
access application object from Sheet or Range (issue #253)

sdementen:
still need to update documentation but could not find location of doc string for xl_app and xl_workbook (that are also initialised in the __init__)

### DIFF
--- a/xlwings/main.py
+++ b/xlwings/main.py
@@ -1446,6 +1446,9 @@ class Chart(object):
         # Use current Workbook if none provided
         wkb = kwargs.get('wkb', None)
         self.xl_workbook = Workbook.get_xl_workbook(wkb)
+        self.wkb = Workbook(xl_workbook=self.xl_workbook)
+        self.app = self.wkb.app
+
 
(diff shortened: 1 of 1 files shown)


sdementen:
For sake of consistency, maybe better to rename .wkb in .workbook (as in xl_workbook)

fzumstein:
I agree on calling it `workbook`. However - this PR suffers from the same bug as issue #236. #277 might be resolving this though - so basically this has to wait until after #277 has been implemented.

sdementen:
ok, let us wait.
for the application object, would this then be self.application (instead of self.app) ?

fzumstein:
I guess that would make sense, yes

sdementen:
Would making these attributes properties that would build the App or Wkb object

sdementen:
Would making the attributes properties that build the object on the fly (by using the Excel COM objects) solve the issue ?

fzumstein:
Don't know, but #277 is at the core of the issue, so I think it should be solved there.

Issue #262, opened 2015-10-18
Range.resize is wrong in case row_size or column_size is None


sdementen:
I would expect that

    Range("A1:C5").resize(row_size=1)
would give a range covering "A1:C1" but it gives the range "A1" (ie if column_size=None, it is assumed to be = 1).
In the code of the resize method, the "if row_size" is maybe better covered by a "if row_size is not None" to catch the case where the size=0 (which should be an error I guess) and ditto for the column_size

fzumstein:
Good catch. It's a plain bug, should work the same as in VBA.

fzumstein:
Should have been closed with 0.5.0

Issue #255, opened 2015-10-16
writing/reading pandas dataframes


sdementen:
It could be useful to have a simple way to read/write dataframes from excel with an API similar to:

    df = Range("A1").df(index=True, header=False).value

or

    Range("A1").df(index=True, header=False).value = df

This could be achieved by having some asdataframe keyword argument for the current Range (similar to asarray) as

    Range("A1", asdataframe=True,index=True, header=False).value = df

However, I wonder if it is not better to leave out the Range object as wrapper of the excel Range object and use accessor to read/write to the range object as

    df = Range("A1").df(index=True, header=False).value
    arr = Range("A1").array(atleast2d=True).value
    lst = Range("A1").list(atleast2d=True).value

(the default being list with atleast2d =False) and the same to write data to it (but there atleast2d would not be required as it can be guessed from the input).
For pandas dataframes:
- instead of df, dataframe, data_frame or DataFrame (as in pandas) could be used
- additional arguments could be added (like index_col, parse_cols, etc as in pandas.read_excel and pandas.to_excel).

We can keep the current syntax for backward compatibility with some WarningDeprecated.

Opinions ?

fzumstein:
Is this only thougt to handle non-hierarchical dataframes?

sdementen:
What are hierarchical dataframes ? multiindex dataframes ?

sdementen:
Regarding atleast2d, I wonder now if it is not better to make it an attribute of the range that:
- can be set manually to True
- is set to True if .table is used 
- is set to True if .horizontal and .vertical are used one after the other
- is set to True if a vertical range is extended with .horizontal (or horizontal range is extended with .vertical)

Then this attribute can be used by the "df" and "array" accessors. The list "accessor" is then not necessary anymore (just use the .value)

fzumstein:
Yes, I was referring to multi-index / multi-header dataframes like these:
```
first              bar                 baz                 foo          
second             one       two       one       two       one       two
first second                                                            
bar   one    -0.410001 -0.078638  0.545952 -1.219217 -1.226825  0.769804
      two    -1.281247 -0.727707 -0.121306 -0.097883  0.695775  0.341734
baz   one     0.959726 -1.110336 -0.619976  0.149748 -0.732339  0.687738
      two     0.176444  0.403310 -0.154951  0.301624 -2.179861 -1.369849
foo   one    -0.954208  1.462696 -1.743161 -0.826591 -0.345352  1.314232
[...]

PR #263, opened 2015-10-18
Implement easy Range <-> dataframe and array conversion #255

sdementen:
Here is a first attempt at accessors. It is unfinished work but I'll be glad to get some feedback on it.
(I have also fixed the resize issue mentionned in issue #262

### DIFF
--- a/docs/datastructures.rst
+++ b/docs/datastructures.rst
@@ -88,14 +88,14 @@ NumPy Arrays
 ------------
 
 NumPy arrays work similar to nested lists. However, empty cells are represented by ``nan`` instead of
-``None``. If you want to read in a Range as array, set the ``asarray`` keyword to True:
+``None``. If you want to read or write in a Range as array, use the ``array`` method of the ``Range`` object:
 
 .. code-block:: python
 
     >>> import numpy as np
     >>> wb = Workbook()
-    >>> Range('A1').value = np.eye(5)
-    >>> Range('A1', asarray=True).table.value
+    >>> Range('A1').array().value = np.eye(5) # Range('A1').value = np.eye(5) also works and calls array() internally
+    >>> Range('A1').table.array().value
     array([[ 1.,  0.,  0.,  0.,  0.],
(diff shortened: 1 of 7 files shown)


fzumstein:
Ideally, the fix for #262 would be submitted in a separate PR, just because the other parts of this PR might take much longer to merge than this simple bug fix. Also, the unit tests should be extended to cover this edge case when either row_size or column_size is None.

sdementen:
I have fix some bugs to pass all tests and remove the fix on issue #262 from this PR.
Still need some doc + extra test for extra functionnality

sdementen:
Hello Felix, could you have a look at this PR ? It is now in more usable form. 
I would need to add a test for a multi-header and multi-index dataframe to confirm it works also for this case.

fzumstein:
sure but you might need to wait a few more days until I find enough time to properly do it. One thing I don't like is the syntax of `as_array` because of [this](http://docs.scipy.org/doc/numpy/reference/generated/numpy.asarray.html). So maybe just `array` and `dataframe` or `df`will be better.

sdementen:
No problem. For the syntax, no strong opinion. I wrote as_xxx because for
list it would have given .as_list instead of just .list that looked weird.
But at the end, the as_list equivalent is the default for Range.value so
not a real problem. I'll change to array and dataframe (maybe more
readable/obvious than df when rereading code)

sdementen:
I wonder if the API could not be simplified to 

    rng.as_array = an array
    rng.as_df = a dataframe

with arguments regarding both passed as attribute of the range (in fact, what is done today).  In this way, as_array and as_df would be just properties of the Range. I would rather keep the as_ (or just as) prefix to avoid using array as a method name (as it is also the array constructor) and I do not see immediately the issue with the asarray method from numpy. It would be similar to the tolist used in pandas but bidirectional (tolist only converts to list not vice versa)

sdementen:
This PR is superseeded by the PR #279 that has been for which I pull from your master. I close it

PR #268, opened 2015-10-20
get_comment & set_comment

geek-ragazza:
get_comment & set_comment

### DIFF
--- a/xlwings/_xlwindows.py
+++ b/xlwings/_xlwindows.py
@@ -387,7 +387,21 @@ def get_formula(xl_range):
 
 def set_formula(xl_range, value):
     xl_range.Formula = value
+    
+def get_comment(xl_range):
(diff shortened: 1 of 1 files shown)


fzumstein:
There's already an open PR #238  for that...

geek-ragazza:
Thank you ! I suggest that add decode ('utf-8') before value write into comment.
Because if you use Chinese without decode utf-8.
Chinese will become garbled.
value=(value).decode('utf-8')
xl_range.Comment.Text(value)

fzumstein:
OK, I ll look into that. But for standard Cell values as in `Range('A1').value = u'some Chinese'` it works correctly, right?

geek-ragazza:
Yes, it works correctly :+1:  Thank you

PR #273, opened 2015-10-22
Issue 269 find window ex raises exce

sdementen:


fzumstein:
`import win32api` is part of the hack, reintroduced here: 6a87a3d77d9bccd567e0c24f5ca8e160f15f3f52

Issue #272, opened 2015-10-22
iterating over a range only works once


sdementen:
I'll add a test and fix it in a next PR

fzumstein:
fixed by 292a1a9fcd50afc8e6e110f3c3e06ecdcb959396

PR #275, opened 2015-10-22
fix iterator issue #272 and add test

sdementen:


fzumstein:
Fails on Python 3: `AttributeError: module 'itertools' has no attribute 'imap'`

sdementen:
I have done the fix for python 3, tests run OK on python 2.7

fzumstein:
that's not enough to make it work across versions, but I've fixed it now: 4b72041c0150315f30bd7acd81feadb4e83a3a08

sdementen:
indeed, I'll try to setup of py3 test environment to avoid such issues in the future

Issue #177, opened 2015-03-29
Implement Sheet protect/unprotect


fzumstein:

fzumstein:
see #284

fzumstein:
Closing this in favor of https://github.com/xlwings/xlwings/issues/1845

PR #284, opened 2015-11-03
Sheet, Range protect and unprotect

geek-ragazza:
Reset my fork and create another branch. Hope this PR will be ok.

### DIFF
--- a/xlwings/main.py
+++ b/xlwings/main.py
@@ -466,7 +466,23 @@ class Sheet(object):
     @name.setter
     def name(self, value):
         xlplatform.set_worksheet_name(self.xl_sheet, value)
+    
+    @property
+    def protect(self):
+        xlplatform.sheet_protect(self.xl_sheet)
 
+    @protect.setter
+    def protect(self, value):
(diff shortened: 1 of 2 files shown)


xlwings/main.py
```diff
@@ -466,7 +466,23 @@ def name(self):
     @name.setter
     def name(self, value):
         xlplatform.set_worksheet_name(self.xl_sheet, value)
+    
+    @property
```
gdementen: The usage of a getter property for doing something other than retrieving a value seems odd/unusual/surprising. I would rather have a normal method with an optional password argument, so that users can call:

```python
sheet.protect()
sheet.unprotect()
sheet.protect('mypassword')
sheet.unprotect('mypassword')
```

OR, if you really want to use a property, you could have a property for being protected without password but you still need a method to unprotect, so the API would not feel very consistent: 

```python
sheet.protect = True
sheet.protect = False
sheet.password_protect('mypassword')
sheet.password_unprotect('mypassword')
```

fzumstein:
closing as this is pre 0.9 refactor. there's an open issue though, see #177

PR #286, opened 2015-11-08
script to create xla and xlam addins

sdementen:
I have no clue how to do it on Mac ...

### DIFF
--- a/build_addins.py
+++ b/build_addins.py
@@ -1,0 +1,32 @@
+import os
+from xlwings import Workbook
+
+
+def build_addins():
+    # transform code for addin use
+    with open(r"xlwings\xlwings.bas", "r") as original, open(r".\xlwings.bas", "w") as for_addin:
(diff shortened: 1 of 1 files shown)


sdementen:
What could also be useful to add on an addin:

 - possibility to specify the different SETTINGS variables (PYTHONPATH, PYTHON_WIN, ...) from the running workbook. We can have some convention that if a named range called PYTHONPATH exists in the ActiveWorkbook, it is used to get back the PYTHONPATH (ditto for the others).
 - possibility to have a form in the addin that can be used to give feedback to the user when a macro is running
 - possibility to add some additional menu items related to xlwings (restart COM server in case of optimised connection ,...)

other ideas ?

fzumstein:
I've refactored things a bit: 20161a8c2c51e47426b2d10075f42227fb3acc04

Also, let's continue the discussion on #301 - I will put in my thoughts a littler later today hopefully.

fzumstein:
Added a few points now on the add-in issue: #301

PR #302, opened 2015-11-19
Get and set font color

geek-ragazza:
set font color 
Range('A1').fontcolor =(255,0,0)

get font color
Range('A1').fontcolor


fzumstein:
I think we'll need to be closer to the original object model (https://msdn.microsoft.com/EN-US/library/office/ff840959.aspx), i.e. make the property return a font object.

fzumstein:
closing as this is pre 0.9 refactor. Replaced with issue, see #897

PR #357, opened 2016-02-02
Accessors experimental

ericremoreynolds:
Experimental pipeline-based approach to accessors logic. Now sufficiently proven to merge in.


sdementen:
Great news Eric ! 
Would you have examples to illustrate how it works ? 
A quick read through the diff of xlwings/tests/test_xlwings.py did not give me a lot of examples.

ericremoreynolds:
Hi Sebastien!

No you're right, you can't see much from the tests.

Basically what's new is the `conversion` module which provides a framework for both standard and custom range read/write functionality. It is in part inspired by your PR, but I redeveloped it because part of the objective is to use the same system for both classic xlwings and the new UDFs.

Perhaps some time we can take a look together as I'd be interested in knowing what you think before releasing it. In the meantime, if you want to get a preview, I suggest you take a look at [the conversion package in the accessors branch](https://github.com/ZoomerAnalytics/xlwings/tree/accessors/xlwings/conversion) where the accessors for standard types are implemented, for example the [DataFrame accessor](https://github.com/ZoomerAnalytics/xlwings/tree/accessors/xlwings/conversion) (which I believe doesn't yet work 100% but you can get an idea). It is possible to write and (if you want) register custom accessors in just the same way as the standard ones.

Say hello to Brussels for me!

Eric

PR #358, opened 2016-02-03
Expand environmental vars

y2kbugger:
Expand the variables in the xlwings.bas config paths.
Use python for PYTHONPATH and UDF_PATH, use the windows shell for the logfile on windows.

This also fixes the slashes in some paths as well, with seeming no bad effects.

Has only been tested on windows and python 2.7.11

### DIFF
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -91,7 +91,8 @@ def xlarg(arg, marshal=None, ndim=None, **kwargs):
 
 udf_scripts = {}
 def udf_script(filename):
-    filename = filename.lower()
+    filename = os.path.expandvars(filename)
+    filename = os.path.normcase(filename)
     mtime = os.path.getmtime(filename)
     if filename in udf_scripts:
         mtime2, vars = udf_scripts[filename]
--- a/xlwings/xlwings.bas
+++ b/xlwings/xlwings.bas
@@ -239,7 +239,7 @@ Sub ExecuteWindows(IsFrozen As Boolean, PythonCommand As String, PYTHON_WIN As S
     WORKBOOK_FULLNAME = ThisWorkbook.FullName
 
     If IsFrozen = False Then
-        RunCommand = "python -c ""import sys; sys.path.extend(r'" & PYTHONPATH & "'.split(';')); " & PythonCommand & """ "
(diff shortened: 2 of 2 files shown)


fzumstein:
Can you quickly describe what the issues are that this PR solves?

y2kbugger:
When deploying, it is helpful that paths can contain environmental variables.
Before, we had to do this, different for each computer:
```
    UDF_PATH = "C:\Users\karlsmith\data_tools/xlwings/ldcUDFs.py"
    LOG_FILE = "C:\Users\karlsmith\data_tools/xlwings/xlwings.log"
```
Now paths can be expanded:
```
    UDF_PATH = "%USERPROFILE%/data_tools/xlwings/ldcUDFs.py"
    LOG_FILE = "%USERPROFILE%/data_tools/xlwings/xlwings.log"
```

y2kbugger:
Also:

> ' For cross-platform compatibility, use backslashes in relative directories

did you mean:

>' For cross-platform compatibility, use _forward_ slashes (/) in relative directories

fzumstein:
Isn't it easier to just use `Environ("USERPROFILE")` on the side of VBA?

The backslash was by intention because of [this](https://github.com/ZoomerAnalytics/xlwings/blob/master/xlwings/xlwings.bas#L349), but you're right, Windows can cope with forward slashes so it should actually not matter which one you use at all.

y2kbugger:
`Environ("USERPROFILE")` would solve my specific individual issue, but I wanted to give back a solution that followed the "rule of least astonishment". It actually took while to realize that the environmental variables were not getting interpreted. This was in part because that issue was confounded by another bug that I forgot to mention:

In the `PYTHONPATH` variable if you are using `"\"`, you need to escape it like this: `"\\"`. This only applied to `PYTHONPATH` the others were working with single `"\"`. This is fixed this using `os.path.normcase()`. Also, `os.path.normcase()` is required because `os.path.expandvars()` returns single `"\"` which is why I forgot to mention this bug.

fzumstein:
OK fair enough - good catch with the PYTHONPATH!

y2kbugger:
I didn't make any changes to `ExecuteMac/ExecuteMac2011` regarding the
environmental variable because I have no way to test that but I imagine
that someone might want use similar variables on OSX.

fzumstein:
true and that's probably also a good argument agains `Environ()` as they don't work on Mac...

PR #378, opened 2016-03-02
Return shortened exception in ud fs

sdementen:
Felix, here is the PR to simplify the returned value for UDFs in case of exception (do not return the full stacktrace). It is still possible to get full stack trace by changing a constant in the ImportPythonUDFs.

PS: I haven't changed doc/.xlam/other files that may need to be changed too

### DIFF
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -117,7 +117,7 @@ def call_udf(script_name, func_name, args, this_workbook):
     return conversion.write(ret, None, ret_info)
 
 
-def generate_vba_wrapper(script_vars, f):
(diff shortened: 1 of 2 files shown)


fzumstein:
Is there a way for you to edit this PR to go against master? Otherwise can you close and open a new one against master (accessors branch is now merged)?

sdementen:
Ok. I will do that and will add another option to allow printing in VBA the time taken by an udf (useful for debugging purposes)

fzumstein:
as discussed...closing this for now so I can get rid of the accessors branch.

PR #387, opened 2016-03-09
Vba timing and short exception message

sdementen:
added possible to define options when generating UDF wrappers + 2 options (timing + short exception message)


skasi7:
My approach with "short exception message" is to have a config option "debug", when is true, all UDFs will return full tracebacks, otherwise, just the exception. This option will be false by default.

For me, the only purpose of full tracebacks is to debug, but once debugged (AKA production), UDFs will return just the exception (more user friendly).

PR #388, opened 2016-03-09
add Workbook accessor

sdementen:


fzumstein:
hey isn't it easier to just do this?

    @xw.func
    def myfunction():
        wb = xw.Workbook.caller()

sdementen:
Ok I did not know the caller ()function for workbooks.

fzumstein:
I ll improve the docs: #400

Issue #77, opened 2014-09-28
dumping large pandas dataframe (~ >10k rows) times out 


notacoat:
a check to see how large the array is and then some code that chunks it into <=10k sections would work.  I wrote a function that does that and it is working fine.

Zenadix:
The function posted by EmilyBarbour is bugged and causes empty rows to appear between chunks. 

I modified the function and now it works fine. I posted on StackOverflow, but I'm posting it here too for easy access.

    # Dumps a large DataFrame in Excel via xlwings
    # Does not include headers.
    def dump_largeDF(df, startcell='A1', chunk_size=100000):
        if len(df) <= (chunk_size + 1):
            Range(startcell, index=False, header=False).value = df
        else: # Chunk df and and dump each
            c = re.match(r"([a-z]+)([0-9]+)", startcell, re.I) # A1
            row = c.group(1) # A
            col = int(c.group(2)) # 1
            for chunk in (df[rw:rw + chunk_size] for rw in
                      range(0, len(df), chunk_size)):
                print("Dumping chunk in %s%s" %(row, col))
                Range(row + str(col), index=False, header=False).value = chunk
                col += chunk_size

For me, a chunk size of 100k is fine, however you may change it depending on your needs.

fzumstein:
Great thanks! Still hope it'll make it into one of the next releases, but as pointed out above, we'll need to implement this further down to cover numpy arrays and lists also.

aschleg:
Thanks a lot for the function Emily and Zenadix, really helpful for moving large datasets between Excel and Python.

I wanted to share a slight modification of Zenadix's function that copies in the column headers as well. Seems to be working well for my purposes so thought it might help others as well :)

```
def dump_largeDF(df, sheet, startcell, chunk_size = 10000):
    if len(df) <= (chunk_size + 1):
        Range(sheet, startcell, index = False, header = True).value = df
    else:
        Range(sheet, startcell, index = False).value = list(df.columns)
        c = re.match(r"([a-z]+)([0-9]+)", startcell[0] + str(int(startcell[1]) + 1), re.I)
        row = c.group(1)
        col = int(c.group(2))
        
        for chunk in (df[rw:rw + chunk_size] for rw in 
                      range(0, len(df), chunk_size)):
            Range(sheet, row + str(col), index = False, header = False).value = chunk
            col += chunk_size
```

fzumstein:
[...]

PR #394, opened 2016-03-11
allow to read large ranges of values (above 4Mio elements)

sdementen:
if the range is above  4Mio elements, the range.Value raises an exception 'Not enough storage is available to complete this operation.'
To go beyond 4mio, we read the data in batches of max 4mio and aggregate the results.

We return lists when range has more than one cell (instead of tuples) as pandas.DataFrame need lists (and not tuples) in their data arg of constructor (useful if we want to be able to skip any "cleaning data" step afterwards in the pipe)

### DIFF
--- a/xlwings/_xlwindows.py
+++ b/xlwings/_xlwindows.py
@@ -1,7 +1,5 @@
 import os
 import sys
-import itertools
-
 
 # Hack to find pythoncom.dll - needed for some distribution/setups (includes seemingly unused import win32api)
 # E.g. if python is started with the full path outside of the python path, then it almost certainly fails
@@ -42,7 +40,6 @@ from . import PY3
 time_types = (dt.date, dt.datetime, type(pywintypes.Time(0)))
 if np:
     time_types = time_types + (np.datetime64,)
-
 
 # Constants
 OBJID_NATIVEOM = -16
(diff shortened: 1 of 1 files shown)


fzumstein:
My first thoughts:

* It should be platform-independent
* I think that the issues is largely memory-dependent, so it shouldn't be a constant but an option: `Range(...).options(chunksize=None).value`

(for both points, see #77)

* I think the list conversion should be part of a (alternative) pipeline stage, but that's Eric's call

sdementen:
I agree with all your points. So better move the chunking in the converter logic. The chunsize can be a general option for some converters but it is probably better expressed as number of elements (vs number of rows... at least it is what I  observed)  
Eric, some hints/directions on how to proceed ?

fzumstein:
I still prefer chunksize=number of rows, because then it'll work the same as e.g. with [pandas.read_csv](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html) and just generally I think it's more intuitive.

sdementen:
Felix, Eric, any evolution on this topic since the time of the issue ?

fzumstein:
shouldn't be a big deal, really ;) I'll add it to v0.10.1...

PR #408, opened 2016-03-18
implemented top-level view() function

gdementen:
I missed a simple view() function, so here is it.

FWIW, I just noticed pyvot also featured such a function.


fzumstein:
manually merged into develop

gdementen:
thanks!

PR #447, opened 2016-04-21
fixed nan for writing pandas in UDFs closes #446

### LINKED ISSUE #446: Nan in pandas
I thought issue 390 was resolved but it seems that it doesn't work if pandas are involved.  I get an #value error in excel upon calling the function below.

```
@xw.func
def test_val():
    a = pd.Series([0,1,2,np.nan,4])
    return a
```

fzumstein:


fzumstein:
closed since targeting master instead of develop

PR #471, opened 2016-05-24
Use importlib.import_module instead of __import__

Juanlu001:
This allows us to programmatically generate a BAS file for a module which is inside of a package. It should be totally backwards-compatible and work on both Python 2 and 3.

### DIFF
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -3,6 +3,7 @@ import re
 import os.path
 import tempfile
 import inspect
+from importlib import import_module
 
 from win32com.client import Dispatch
 
@@ -144,7 +145,7 @@ def get_udf_module(module_name):
             module_info['module'] = module
             return module
     else:
-        module = __import__(module_name)
+        module = import_module(module_name)
         filename = os.path.normcase(module.__file__.lower())
         mtime = os.path.getmtime(filename)
         udf_modules[module_name] = {


Juanlu001:
Pinging @skasi7

Juanlu001:
See for instance a trivial case to show the difference:

```python
>>> __import__("os.path")
<module 'os' from 'C:\Python27\lib\os.pyc'>
>>> from importlib import import_module
>>> import_module("os.path")
<module 'ntpath' from 'C:\Python27\lib\ntpath.pyc'>
```

fzumstein:
it's not compatible with 2.6, but I am not necessarily against dropping 2.6 support...

skasi7:
It could be changed only for Python versions > 2.6, leaving the current code for <= 2.6. This way you have the new features without dropping compatibility.

Juanlu001:
According to some statistics, Python 2.6 is not very popular nowadays:

- https://caremad.io/2015/04/a-year-of-pypi-downloads/
- astrofrog.github.io/blog/2015/05/09/2015-survey-results/

Apart from that, according to PyPI stats almost all the 2.6 downloads come from Linux, so I think it's safe to drop it for Windows and OS X. I found the data using the explanations found here:

https://mail.python.org/pipermail/distutils-sig/2016-May/028986.html

and this query:

https://gist.github.com/alex/4f100a9592b05e9b4d63#which-platforms-are-generating-26-downloads

fzumstein:
extended/merged into develop by #491

PR #521, opened 2016-07-27
fix typos in migrate doc

gdementen:


fzumstein:
First is already fixed (though not pushed yet), 2nd is actually correct: it returns a name object now as in VBA.

PR #589, opened 2016-11-28
UDF arguments documentation is now available

Colin-b:
Also avoid iteration on arguments in case excel version is < 14 (2010).
Also iterate only once over arguments in case excel version is >= 14 (2010).

I am also quite surprised that it was not working before but it fix the issue with my setup, I guess COM call should be completed with all arguments.
Description for each argument can be found on MSDN: https://msdn.microsoft.com/en-us/library/office/ff838997%28v=office.15%29.aspx?f=255&MSPPError=-2147217396


fzumstein:
great! will try to merge on the weekend!

PR #590, opened 2016-12-05
Allow to specify a category for each UDF

Colin-b:
An integer can be provided (between 1 and 14 included)
A string can be provided to create a new user defined category.

Default behavior stay the same, instead of providing None as the Category, 'User Defined' category 14 is explicitely provided in COM call.

More information on the meaning of each integer value can be found on MSDN: https://msdn.microsoft.com/en-us/library/office/ff838997%28v=office.15%29.aspx?f=255&MSPPError=-2147217396

I cannot find the test cases related to UDFs, could you please point it to me so that I can add new ones?


fzumstein:
The tests are here: https://github.com/ZoomerAnalytics/xlwings/tree/master/xlwings/tests/udfs under udf_tests.py/xlsm

fzumstein:
Should I move forward with this PR or were you going to add tests first?

Colin-b:
I did not had a look at test cases yet but I will (not before January 2017 sadly). I am pretty confident of this modification as I am using it on my forked version and I did not encounter any issue with it but it is always better to have some kind of reliable test suite asserting the behavior ;)

fzumstein:
fair enough thanks

Colin-b:
I added some test cases, you can find the new test cases at the end of the XLSM file. Copy/Pasting them at the end of the current version should fix the conflict.

I am not really sure I checked errors the best possible way in the test python file (I kinda dislike the try/catch/fail antipattern), if you have any proposal to do it in a more elegant way I would love to hear it.

PR #591, opened 2016-12-05
Handle VBA generated lines with more than 1024 characters

Colin-b:
Excel already try to do it but it remove the last character of the line by doing so


Colin-b:
I added some test cases, to solve the conflict you can copy/paste the last 2 tests on the XLSM file.

PR #606, opened 2017-01-12
Avoid calling UDF with wizard

Colin-b:
Add the ability to avoid calling an UDF when the Excel Function wizard (allowing to specify arguments) is opened.
This is particularly useful when an UDF does not return instantly.
Use `writeln` method instead of `write` to ease code understanding.


fzumstein:
I refactored `call_while_in_wizard` into `call_in_wizard`, i hope that's ok with you

Colin-b:
That's perfect :)

PR #610, opened 2017-01-16
Retry COM call in case retry error is returned

Colin-b:
Retry COM call in case RPC_E_SERVERCALL_RETRYLATER error is returned (arbitrary set maximum number of retry to 9, total number of trial being 10 then)

For more information refer to https://msdn.microsoft.com/en-us/library/ms228772(v=vs.100).aspx?f=255&MSPPError=-2147217396

I am not fond of the fact that the number of retry is hardcoded here so if you find a way of providing options to xlwings properly it would be nice to introduce a property I guess.
Also please note that 10 might be a number a bit too high, in my case, most of the time 2 or 3 attempts were enough.

### DIFF
--- a/xlwings/server.py
+++ b/xlwings/server.py
@@ -329,12 +329,29 @@ def serve(clsid="{506e67c3-55b5-48c3-a035-eed5deea7d6d}"):
                 break
 
             task = idle_queue.pop(0)
-            try:
-                task()
-            except:
-                import traceback
-                print("TaskQueue '%s' threw an exception: %s" % task, traceback.format_exc())
+            _execute_task(task)
 
     pythoncom.CoRevokeClassObject(revokeId)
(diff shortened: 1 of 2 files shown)


fzumstein:
Can you provide a sample issue that this resolves?

Colin-b:
It's pretty hard to reproduce in a deterministic manner but the file was containing multiple sheets, each sheet with a different UDF call and some parameters always shared between the UDFs.

Meaning that changing the value of a parameter (in a specific cell, not directly as a parameter), results in calling every UDFs on the file.
When sending back the result (code section that is affected by this PR), sometimes the issue does not occurs, sometimes it does.
I cannot share my file but those are large (~10 columns and many thousand rows) dynamic arrays sent as results (and computation might take seconds to perform before returning result).

PR #611, opened 2017-01-16
Always reload module on import UDFs

Colin-b:
Some modules might requires code to be executed to regenerate UDFs for example.

### DIFF
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -180,28 +180,28 @@ def get_udf_module(module_name):
     module_info = udf_modules.get(module_name, None)
     if module_info is not None:
         mtime = os.path.getmtime(module_info['filename'])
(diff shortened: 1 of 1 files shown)


ericremoreynolds:
Hi Colin,

Regarding this PR we'd be more in favour of adding a "Restart Python" button on the add-in. Module reloading does not solve the problem of picking up code changes in general (e.g. if the changes are in a module used by the UDF module). Restarting Python would be cleaner, and there could be an option to trigger it every time the UDFs are imported.

Unless there is some other use case I haven't thought of?

Thanks, kind regards,

Eric

Colin-b:
That's fine for me, this is way better obviously.
Thanks

fzumstein:
replaced by #630

PR #630, opened 2017-02-19
Kill python process

fzumstein:


Colin-b:
Hi,

It seems related to https://github.com/ZoomerAnalytics/xlwings/pull/611#issuecomment-273096000

As there is no details about this on the changelog is it considered as stable?

Kind Regards

fzumstein:
Yes, it works but we will release it as part of the new runtime addin.

PR #650, opened 2017-04-05
fix #595

### LINKED ISSUE #595: Range.__iter__ iterates wrong cells
with code

    for cell in xlwings.Range('A1:A3, C2:C3, D4'):
        print(cell.address)

i got

    $A$1
    $A$2
    $A$3
    $A$4
    $A$5
    $A$6
------------
which need to be

    for cell in xw.Range('A1:A3, C2:C3, D4').impl.xl.cells:
        print(cell.address)

    $A$1
    $A$2
    $A$3
    $C$2
    $C$3
    $D$4
------------
Sorry about my not so excellent English.

akasolace:


fzumstein:
Mac implementation is pending here. Quick note: 

```
for i in self.api.cells.get():
    yield sheet.range(i.get_address())
```

fzumstein:
The other issue is that if we add support for multi regions then we need to make them work across the board, not just when iterating over it.

akasolace:
@fzumstein that PR is not really a new feature but just a bug fix.
At the moment when discountinued selection is made in excel ... the iteration is broken and this PR solely intended to fix that particular bug (fyi it is a hack I apply on xlwings since couple of month before distribution of my code)

akasolace:
I agree this hack does not make xlwings able to work with discontinued selection properly so it does not close #595 as it claimed however I believe it is a small step in that direction.  
The problem is that if a user try to use discontinued selection, he can but the result would be incorrect... maybe until there is a full support, an error message or at least a warning should be thrown when trying to set values and or to iterated over discontinued selection.

akasolace:
I was wondering if multiple range support is still being considered ?
Still have to do this hack (change those lines in main.py of all users) everytime we change version of xlwings. I am doing it with a batch file and some regex, so it not a big deal, but still not really clean.

fzumstein:
What version are you on? Anything like 

```
for cell in wb.sheets[0].range('A1:A3, C2:C3'):
    print(cell.address)
```

throws an exception with the latest version (as it should) so an easy solution would be to change your application code rather than hacking the xlwings source code I guess...

akasolace:
I just upgraded from 0.11.4 to 0.12.1 and for the time being this hack still work
I agree with you that hacking the main.py is not really clean, it is why I was asking if eventually support would come. If not I will indeed change my application code and plug some VBA in the application code.

fzumstein:
Can you add a mini sample of how you are using it? I currently don't understand why you can't use multiple xlwings Range objects (i.e. I dont get the necessity of plugging VBA code).

akasolace:
@fzumstein , 
just in case you missed it ... I replied to your last question by editing my previous comment

fzumstein:
OK, at least I see where you're coming from now. But you can't use `api` here as this breaks cross platform compatibility. the `Range.cells` property would have to be implemented within both `_xlmac.py` and `_xlwindows.py`.

The "proper" workaround would be to fall back to the pywin32 object:

```
import xlwings as xw
wb = xw.books.active
for cell in wb.app.api.Selection:
    print(cell.GetAddress())
```

akasolace:
ah ok, I will try this then.
Thank you

fzumstein:
Ok cool, I am closing this in the meantime, but feel free to ask more questions about the workaround.

akasolace:
I confirm the workaround is working and solve my issue.
Thank you !!

PR #685, opened 2017-06-21
Avoid RPC_E_SERVERCALL_RETRYLATER failure

Colin-b:
When multiple UDFs are depending on the same cell, when this cell is updated all those UDFs will be called by Microsoft Excel.
It seems that Microsoft Excel will ask for a retry until all UDFs have been called.

In order to avoid failure in case this maximum number of retry is reached and all UDFs have still not been properly called, retry should be performed only after last call (the first one that will not be rejected with a retry issue)

### DIFF
--- a/xlwings/server.py
+++ b/xlwings/server.py
@@ -338,17 +338,33 @@ def serve(clsid="{506e67c3-55b5-48c3-a035-eed5deea7d6d}"):
 def _execute_task(task):
     try:
         task()
+        logging.debug("Task '{0}' was properly sent to Microsoft Excel.".format(task))
+        for retry_task in retry_queue:
+            # If a retry failed it means that there is still call to be performed
+            # (It should really never occurs as the task triggering those retry was not rejected).
+            if not _retry_task(retry_task):
+                return
+        retry_queue[:] = []
     except Exception as e:
-        if _ask_for_retry(e) and _can_retry(task):
-            print("Retrying TaskQueue '%s'." % task)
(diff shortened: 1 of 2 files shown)


Colin-b:
Hi,

Is this issue fixed in the new 0.11.0 release?

Thanks and Regards

fzumstein:
Sorry about that, must have been by accident. Ah, now I see why: I deleted the develop branch.

Can you open a new one against master? However, we're kind of in the dark here and don't really get what you're doing. This kind of applies also to your previous (merged) PR: https://github.com/ZoomerAnalytics/xlwings/pull/610

Could you either comment/explain a bit better what you're doing and/or attach a sample file/test case so that we can replicated what exactly it's doing?

thanks

Colin-b:
Sure no problem, I'll do it tomorrow and I will add a test case reproducing
the issue.

Best Regards

Colin-b:
I'll do a new PR once I have merged and validated my changes with the new xlwings version.

In the meantime if you want to reproduce it you can do it pretty easily by setting A1 to 1 and calling the following UDF 100000 times in an excel file (=auto_expanded_table(A1)).

If you now switch A1 value to 2, all UDFs should update and some will not be updated (with the actual xlwings version), with this fix all values will be properly updated

```
@xw.func
@xw.ret(expand='table')
def auto_expanded_table(input_param):
    return [['header 0', 'header 1'],[input_param, input_param]]
```

PR #780, opened 2017-11-20
Enable cross-platform development of UDFs

rbubley:
Currently, if you have a file that contains UDFs destined for windows, with a decorator like: `@xw.arg('foo', doc='bar')`, this would generate an error on a non-windows platform:

```TypeError: arg() got an unexpected keyword argument 'doc'```

### DIFF
--- a/xlwings/__init__.py
+++ b/xlwings/__init__.py
@@ -44,8 +44,10 @@ else:
     def sub(f):
         return f
 
-    def ret(f):
-        return f
+    def ret(*args, **kwargs):
+        def real_decorator(f):
+            return f
+        return real_decorator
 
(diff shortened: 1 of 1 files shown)


fzumstein:
this should be fixed here: https://github.com/ZoomerAnalytics/xlwings/blob/master/xlwings/__init__.py#L37-L51

rbubley:
Hi @fzumstein, the existing code doesn't work if arguments are passed to these decorators. The PR amends the lines you indicated to swallow decorator arguments on non-windows machines.

fzumstein:
sorry that was too quick! could you please do the same for func and sub, then I can merge, thanks!

PR #798, opened 2018-01-18
WriteValueToRangeStage construction error

skasi7:
RawValueAccessor is not providing the options argument to the WriteValueToRangeStage constructor.


fzumstein:
do you have a sample that gets fixed by this? thanks

skasi7:
Just use @xw.ret(raw=True) in an UDF. Is broken in 0.11.5. This change fixed the issue.

skasi7:
Thank you.

Issue #787, opened 2017-12-15
multiple excel workbooks with UDF cause multiple COM servers if placed in different folders


PJCampi:
Let's say I have the following folder structure:

. root_folder:
____- wbk_1.xlsm
____. subfolder:
________- wbk_2.xlsm

Let's assume that wbk_1 and wbk_2 each have one xlwings UDF function and a reference to the same xlwings.xlam.

If I run both function in the same excel application, two python processes will be created. I expected only one shared process for both workbooks. If they are in the same folder, only one process is created.

I run v11.4 on windows with a single virtual environment and I have added _root_ and _subfolder_ to the path configuration.

Is that expected behaviour of xlwings? Is there a workaround to hook the workbook to the same COM server?

skasi7:
Different command lines lead to different processes (I assume there's some kind of cache involved). And since PYTHONPATH is part of every command line and always starts with ActiveWorkbook.path, each workbook will spam a new process. I don't know if this is intended. This behavior can be changed modifying the addin.

PJCampi:
Hey skasi7,

I have tested your suggestion and it works.

The activation or creation of a COM server is a dependent on the command passed to the C++ function _XLPyDLLActivateAuto_.

As you point out, the question now is whether the current behaviour is intended or should be changed...

PJCampi:
For reference, I have made the following change to the xla for my specific use case:
- the python command arguments are changed to only register the path of the xla when the python process is created
- I have added a UDF to the xla that registers the path supplied to the python path
- all my spreadsheets must refer to the xla and register themself to the python path via the aforementioned xla UDF
works fine for my purpose...

fzumstein:
Can you upload your addin somehow? I believe it's blocked here in the issues, but you could do a PR or post a link to a fileshare platform.

PJCampi:
I have made very little changes so maybe it's probably easiest to post them here:

**Main.XLPyCommand:**
    _-> initial code:_ 
                `PYTHONPATH = ActiveWorkbook.path & ";" & GetConfig("PYTHONPATH")`
    _-> new code:_ 
                 `PYTHONPATH = ThisWorkbook.path & ";" & GetConfig("PYTHONPATH")`

**xlwings_udfs.py:**
    _-> new code (func to be registered in the xlam):_
```
import sys
import xlwings as xw

@xw.func
def py_add_to_path(path):
    sys.path.append(path)
```

**workbook to register:**
    _-> new code:_
```
[...]

Issue #781, opened 2017-11-28
Catching Worksheet events


cihant:
Hi,

I want run python code when Worksheet_Calculate or Worksheet_Change event fired. How I can? 

Thanks

skasi7:
I think you can't as of the current release (v0.11.5). I have an implementation of this feature and I've commented it on #555. @fzumstein Do you prefer to use this to discuss about event handling and #555 for object cache or to concentrate both in #555.

fzumstein:
yeah maybe this issue for event handling, but whatever is easier

skasi7:
I think I'll post here about event handling and when events are accepted, we can move on to object cache in #555 (which needs event handling anyway).

skasi7:
Ok, so the implementation goes like this:
1. Needs some code inside the ThisWorkbook portion of the addin to catch Excel events and propagate them to Python, like this (this is only an example, but every Excel event can be handled):
```vba
Private Sub App_NewWorkbook(ByVal Wb As Workbook)
    Py.Event "NewWorkbook", Array(Wb), ActiveWorkbook, Application.Caller
End Sub
```
2. A new Event method inside XLPython class will receive the event notification and send it to an event dispatcher.
3. User code registers in the event dispatcher to receive the event notifications, like this:
```py
EventDispatcher.register_callback(event_name, callback)
```
4. This can be easily arranged as a UDF decorator, like this:
```py
def xlevent(event_name):
    def inner(f):
        EventDispatcher.register_callback(event_name, f)
        return f
    return inner
```
This can be use like this:
```py
@xlevent('NewWorkbook')
def LogNewWorkbook(xl_workbook):
    _logger.debug(u"[Event] New workbook '{}'".format(xl_workbook.FullName))
```
5. There's also support for a callback in case of exception of an event handling, useful for logging.
6. An for the OO lovers, a metaclass is also created to be able to have things like:
```py

class EventLogger(object):
    __metaclass__ = xlwings.EventHandler

    @classmethod
    def NewWorkbook(cls, xl_workbook):
        _logger.debug(u"[Event] New workbook '{}'".format(xl_workbook.FullName))
```
7. As you can see for the previous example, there's no wrapping of the COM objects to xlwings, is something I was thinking about it, but not implemented it.

What do you guys think? This implementation matches your current goals of the library? I'm currently working into the port of the implementation to v0.11.5 (I've been working in v0.6.4 since the beginning and I'm now migrating into all the new stuff).

[...]

PR #802, opened 2018-01-29
First approach to events.

skasi7:
PR for issue #781.

These are the interesting changes:
* (event_dispatcher.py) Added to handle event dispatching.
* (addin.xlam/Config.bas) Fixed issue with GetWorkbookDirectoryConfigFilePath when ActiveWorkbook is Nothing.
* (addin.xlam/Main.bas) Fixed issue with XLPyCommand when ActiveWorkbook is Nothing.
* (main.py) Added caller_address and caller methods to Range class to retrieve the caller address and caller range.
* (server.py) Alternative kill procedure (win32con.WM_QUIT message). Event method to process events from addin. Disabled events during CallUDF and Event methods to avoid recursive events. New protected method _TraceExc to trace exceptions to log and/or message box.
* (udfs.py) Added xlevent and xleventexc decorators to create event handler UDFs. Removed BOOK_CALLER setup inside call_udf (moved to CallUDF and Event methods from server.py).

Also an example of use were added:
* Steps to reproduce:
  * Open Excel
  * Close default workbook
  * Load event.xlsm
  * Import UDFs (will connect event handlers), multiple imports won't trigger multiple times.
* Things to play with:
  * Click on any, cell => Sets value to X (note that this not triggers the WorksheetChange event)
  * Modify any cell => adds 1 if it's numeric (shallowed exception if not)
  * Change sheet => A message box will show.
  * Close and open events.xlsm (don't work with other workbooks due to #787) => A message box will show (lack of creativity).

TODO:
* Add more event handlers (as needed/requested).

Any comment will be welcomed.


fzumstein:
thanks! It might take me a few days though until I get around to take a closer look...

skasi7:
Feel free to ask me whatever you don't catch (my code can be overcomplicated sometimes).

fzumstein:
finally got to look at this! I think it's great, only two questions for now:

* is it possible to setup a spreadsheet so it connect the handlers automatically when opening it instead of having to click on import udfs? 
* On these samples: 

```
@xw.event('SheetSelectionChange')
def on_sheet_selection_change(xl_sheet, xl_range):
    xl_range.Value = 'X'
```

is it possible to use the xlwings range object here instead of the pywin32 one?

I've pushed your branch here so to see the vba diffs: https://public.xltrail.com/#/workbook/github.com%2FZoomerAnalytics%2Fxlwings.git%2Fxlwings%2Faddin%2Fxlwings.xlam/diff/b92eca84e5d101b7cd7fc894737bf88cd9dcc0f6

skasi7:
Glad to read that you like it! Answering your questions:
* The problem is to know the location of the UDFs module. If it's known, you can add the ImportPythonUDFs macro to the Auto_Open Sub (executed when a spreadsheet is opened manually).
* The reason behind not using xlwings objects was I usually perform simple operations. But it's perfectly possible to replace the native objects with xlwings equivalents in the event_dispatcher.py before calling the user defined callbacks.

Thank you for the diff in xltrail (very cool tool BTW), It's very difficult to see these changes.

fzumstein:
* Regarding the location of UDF module: I don't see the difference to standard UDFs, i.e. you could set the location of the module (i.e. PYTHONPATH & UDF Modules combo) either via global config/ribbon or via `xlwings.conf` sheet if you don't use the default (same name, same directory).

* Is there a reason why you would use Auto_Open instead of Workbook_Open?

* Would you have time to add the changes for the xlwings object? since you can always fall back to `.api` it would be more in line with the rest of the library.

skasi7:
* You're right about the location. The reason behind using Auto_Open is that it's only called when the workbook is opened manually (not programatically, although can be fired if needed). I think it's safer.
* Yeah, sure, I can change the callbacks to pass xlwings objects. I agree is more consistent, but sometimes I'm a little lazy ;).

fzumstein:
Actually, I've setup your fork on xltrail so you can see the diff automatically: https://public.xltrail.com/#/?path=github.com%2Fskasi7%2Fxlwings.git&branch=events

skasi7:
Finally I've managed to take some time to change the arguments of the callbacks from pywin32 instances to xlwings objects. If you have questions, don't hesitate to comment.

skasi7:
My decorators mimic the others from xlwings. The first time I saw them I was also intrigued, but never asked. Maybe Felix can enlighten us.

fzumstein:
well, it's the same as further down where it's called `inner` instead of `real_decorator`: it's used so the decorator can accept args. It maybe confusing because the instance with `real_decorator` are indeed just dummies so that the functions can still be used on Mac.

fzumstein:
Ok so I had another look at it with @ericremoreynolds who did the UDF part. We would like to simplify that to generate the event wrappers only in the workbook itself and only for the ones that are used. This should be doable within the `import_udfs` function and written directly to the workbook using something along these lines (just a sample as a reminder): 

```
Sub Something()
    Workbooks("...").VBProject.VBComponents("ThisWorkbook").CodeModule.AddFromString ("Private Sub Workbook_SheetChange(ByVal Sh As Object, ByVal Target As Range)" + vbCrLf + "Application.StatusBar = Target.Address" + vbCrLf + "End Sub")
End Sub
```
https://www.ozgrid.com/forum/forum/help-forums/excel-general/16556-importing-exported-modules-into-sheet-code-modules-and-thisworkbook-module

skasi7:
Sounds great! The more autogenerated VBA code, the better. And creating only the needed ones is a neat touch.

lehajam:
Hi guys, 

Just would like to know if there have been any update on this as I would like to subscribe to server sent events from a UDF ?

Thanks

fzumstein:
what do you mean by server sent events? This PR is about Excel workbook/sheet events.

richbeales:
Hi, is there any timescale/further progress on this?

fzumstein:
Hi @richbeales, thanks for the bump, we're planning to implement this as part of xlwings PRO in 2020.

Issue #555, opened 2016-08-24
Add caching to return Python objects to Excel cells


fzumstein:

ericremoreynolds:
References: [1](https://zoomeranalytics.uservoice.com/forums/269851-xlwings/suggestions/15495960-support-general-python-object-as-return-type-or-ar) [2] (http://stackoverflow.com/questions/38537566/when-using-xlwings-can-i-return-a-object-to-excel-cell)

skasi7:
I have a implementation of this feature (based on https://www.pyxll.com/docs/examples/objectcache.html but simpler). The problem is this feature requires event handling (that I also implemented) to be efficient.

If you are interested, I can send you more information/examples/PRs.

sdementen:
+1

fzumstein:
sure, please send!

audetto:
This is similar to the issue I recently opened https://github.com/xlwings/xlwings/issues/1541
I work in finance and have see countless poor implementations of this concept which make spreadsheet a nightmare to follow, with issues like: F9 does not work, memory builds up, multiple sheet do not work together.

I think Microsoft should solve it once and for all, but IMHO xlwings is the next best location to do so for python objects.

What I am writing at the moment uses these ideas

- put the object in a global variable (a dictionary)
- keyed by the cell address
- add some visible tweaks to identify as objects, giving the impression that the object has been recomputed

so one could use keys like

object:[filename.xls]Sheet3!B4@10:45:56

where the time is not used in the lookup (because we must delete the object already in the same cell).
optionally, add the type as visual info (not used in lookup)

DataFrame:[filename.xls]Sheet3!B4@10:45:56

Converters are really a good place for this, as they ensure a full separation between object handling and function code (another pitfall of the many bad designs).

sdementen:
You may try the converter `CacheConverter` below that can be used as:
```python
import pandas
import xlwings as xw


@xw.func
@xw.arg("df", pandas.DataFrame)
@xw.ret(CacheConverter)
def write_df2cache(df):
    return dict(my_value=df, my_other_value="45", the_third_value=dict(info=len(df),cols=df.columns))



@xw.func
@xw.arg("df", CacheConverter)
def read_cache(df):
    return df


@xw.func
def see_cached_keys():
    return [[e] for e in _cache_name_timestamp]
```
In Excel, you use `write_df2cache` to read a range. This will return 3 values as reference similar to:
[...]

PR #803, opened 2018-01-29
First approach to object cache.

skasi7:
PR for issue #555.

Note: The branch of this PR is a branch of the events branch, so contains changes from both features. The only file modified in both is __init__.py (importing the new stuff to the main namespace).

These are the interesting changes:
* (object_cache.py) The new Object cache.

Also an example of use were added:
* Steps to reproduce:
  * Open Excel
  * Close default workbook
  * Load object_cache.xlsm
  * Import UDFs (will connect the object cache event handlers),
* Things to play with:
  * See the formulas inside the cells forming the table.
  * Note that the last instance is an aggregation of instances returning another instance.
  * See the amount of objects and references hold by the cache.

TODO:
* Removal of references (currently the only approach to remove references is removing them, closing Excel and start again). This will imply handling more events and updating the object cache accordingly (see https://www.pyxll.com/docs/examples/objectcache.html).

Any comment will be welcomed.


skasi7:
I changed the event handlers inside objects_cache.py to match the new arguments passed (xlwings objects). So this implementation is also ready for review.

fzumstein:
Awesome thanks!

audetto:
I think this would be a great addition to xlwings.
One question I have is:

**Assume, one cannot hook Excel events, how good / bad the rest of the idea would be?**

I see 2 events are intercepted: ``Open`` and ``NewSheet``.

I was surprised by this choice: I thought that one needs to hook the ``Close``, ``Delete`` or ``Calculate``, because they signal when cells die.

So, what would be the consequences of missing ``Open`` and ``New``? 

- a previous xls had registered an object at some cell ``A.xls.Shee1.D1``
- it is closed
- a new one (with the same name) is opened which does not define ``A.xls.Shee1.D1``, but the old object is still available?

Is this the main issue that events try to solve?

I am only commenting this as I wonder if hooking events does not add a whole new complexity to ``xlwings``?

fzumstein:
This has now also been implemented in xlwings Server, see: https://github.com/xlwings/xlwings/issues/555#issuecomment-2289741057

PR #832, opened 2018-03-18
Remove password protection on the xlwings VB addin

Colin-b:
As the license already allow modification, the password protection does not seems to make sense here.
Plus it prevents browsing VBComponents (at least in C#).


Colin-b:
AppVeyor build seems to fail due to this removal as it expect the file from being password protected I guess. It would certainly requires some more modification but I do not feel confident in modifying a build script like this one. Could you handle it?

fzumstein:
The password is basically there to hide the addin away in the VBA explorer so it stays out of sight when working on other things. It also helps VBA unaware users to not accidentally change code when something fails and they click on debug. Not really sure what you do with C# but during development nobody stops you from temporarily removing the password.

As for the AppVeyor build: It always fails for PRs as the build makes use of secure env vars and having these always fails for PRs.

Colin-b:
Those are valid points indeed, preventing users from unwanted modification.
In such a case, couldn't you set the file as password protected only when installing the add-in? 

That way the xlam file could be accessed by programmers without protection if needed (well, this is my use case)

PR #833, opened 2018-03-18
Generate UDFs on ThisWorkbook instead of ActiveWorkbook as in v0.10

Colin-b:
It seems like a regression since the move to the new VB addin.
UDFs were generated using ThisWorkbook, allowing for generated "xlwings_udfs" VB module to be within the calling project instead of the ActiveWorkbook project.


fzumstein:
Currently, the addin generates the wrapper code in the active workbook when you click the import button. Your version would save it in the addin which might make sense in some use cases (like for building custom extensions in the style of sql) but would need more changes than just replacing ActiveWorkbook with ThisWorkbook.

Colin-b:
Well you know better but could you elaborate on those other changes?

As I said this was the behavior on xlwings 0.10.4 (before the brand new add-in) and I had the feeling it was working fine (at least for UDFs). Would't it be possible to use ThisWorkbook only for UDF calls then?

PR #834, opened 2018-03-18
Allow to provide an add-in configuration file path

Colin-b:
This path can be provided on a per call basis, thus every call to XLWings add-in can be performed with a different configuration while keeping the user configuration untouched.

This will allow custom user configuration while allowing other tools to call with a tool specific configuration


fzumstein:
What would be a use case that could not be solved by either the workbook config sheet or the directory config file?

Colin-b:
I have a custom add-in on top of xlwings, this add-in has his own xlwings configuration.
Now if the user also wants to use xlwings, he should be able to but my calls to xlwings should always use my configuration.

fzumstein:
Wouldn't it make more sense that you include xlwings in the form of the single module that you get when you do `xlwings quickstart myproj --standalone` and include that with hardcoded configs in your addin? This way you can also be sure that vba and python versions are in sync.

PR #836, opened 2018-03-19
Xlwings conf (fix #830)

### LINKED ISSUE #830: xlwings quickstart could create xlwings.conf with right interpreter
(with xlwings=0.11.7, windows, excel 16.0)
when using `xlwings quickstart myproject`, it could be nice to have a xlwings.conf created in the folder myproject with the INTERPRETER pointing to the python executable used by xlwings (and maybe remove the _xwlings.conf sheet in the excel project to avoid confusion or adapt the comments in the sheet).

sdementen:


fzumstein:
what about making that optional with `--dirconf` or something?

sdementen:
IYO, which would be the best default, sheet or dir conf?
The issue with sheet version is that you cannot modify it from python without opening excel, which can be tricky in general, so it is difficult to preconfigure with the right python interpreter.
The dir conf may be a bit more abstract as there is one more moving piece but it sounds to me better (a vague feeling)

fzumstein:
well, the current default is the global ribbon setting as the sheet config is disabled. We can think to use a template without the sheet config when `--dirconf` is chosen. But you also have to see that preconfiguring the python interpreter the way you do it may work for you but not for many others (e.g. many who have installed anaconda on the users directory and want to give the projects to coworkers). In that case, being able to find conda or virtualenvs without full path would be much more helpful I think (see https://github.com/ZoomerAnalytics/xlwings/issues/422).

PR #872, opened 2018-05-14
Return UDF result even if 10 return fail

Colin-b:
When result cannot be sent because Excel is busy, wait for next successful call to return result

### DIFF
--- a/xlwings/server.py
+++ b/xlwings/server.py
@@ -25,6 +25,7 @@ import win32com.client
 import win32com.server.util as serverutil
 import win32com.server.dispatcher
 import win32com.server.policy
+import logging
 
 from .udfs import call_udf
 
@@ -277,6 +278,7 @@ class XLPython(object):
         exec (stmt, globals, locals)
 
 
+retry_queue = []
 idle_queue = []
 idle_queue_event = win32event.CreateEvent(None, 0, 0, None)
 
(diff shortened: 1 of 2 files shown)


Colin-b:
Would it be possible to consider merging this while #881 is still under development ? It's the only remaining development I am missing in official xlwings package to be able to release a public version of my package (without having to publish a fork on pypi) :)

fzumstein:
yes getting closer but will need a sample workbook to replicate first, maybe @DougAJ4 Excel file will do: https://newtonexcelbach.com/2018/05/24/more-numpy-with-excel-and-xlwings/

Colin-b:
I created a test file to try to reproduce the issue (result could not be written even if retrying 10 times as Excel is still calling some functions).

The file contains 200 times the same function, all calls are triggered by changing the value in A1, the value corresponds to the number of columns to write as result.
The 11 first columns of B row of the document contains the number of filled cells in respective column.

Steps to reproduce:
 * Change the number in A1 for a value between 1 and 11 (included).
 * Notice that once all calls are performed by Excel, B row cells are not containing 200 (the expected number of rows to be filled)

Here is the Python code:
`import xlwings as xw


@xw.func
@xw.arg('nb_rows', numbers=int)
@xw.ret(expand='table')
def hello(nb_rows):
    return ['value'] * nb_rows
`

The Excel file (and the Python code, for your convenience) is attached as zip.
[xlwingstest.zip](https://github.com/ZoomerAnalytics/xlwings/files/2491592/xlwingstest.zip)

fzumstein:
👍 thx!

fzumstein:
You're absolutely right @DougAJ4, this is the way to go, but it'll probably take a few more years until it's available to the average corporate employee...

fzumstein:
With your branch, I sometimes get this error:

```
ERROR:root:An error occurred while executing task '<xlwings.udfs.DelayWrite object at 0x000001A4D855B400>'.
  File "c:\users\felix\dev\xlwings\xlwings\server.py", line 352, in _execute_task
    task()
  File "c:\users\felix\dev\xlwings\xlwings\udfs.py", line 185, in __call__
    .override(_skip_tl_cells=self.skip)
  File "c:\users\felix\dev\xlwings\xlwings\conversion\__init__.py", line 35, in write
    pipeline(ctx)
  File "c:\users\felix\dev\xlwings\xlwings\conversion\framework.py", line 66, in __call__
    stage(*args, **kwargs)
  File "c:\users\felix\dev\xlwings\xlwings\conversion\standard.py", line 46, in __call__
    if ctx.range and self.expand:
  File "c:\users\felix\dev\xlwings\xlwings\main.py", line 1067, in __len__
    return len(self.impl)
  File "c:\users\felix\dev\xlwings\xlwings\_xlwindows.py", line 685, in __len__
    return (self.xl and self.xl.Count) or 0
  File "C:\Users\felix\Anaconda3\lib\site-packages\win32com\client\dynamic.py", line 527, in __getattr__
    raise AttributeError("%s.%s" % (self._username_, attr))
AttributeError: Range.Count
```

fzumstein:
whereas I can't really find anything wrong with https://github.com/ZoomerAnalytics/xlwings/issues/880

Colin-b:
Do you mean that sometimes results are not sent back to Excel ? I must confess I did not try the sample reproducing current issue using #880 but all results are sent back (at least that's the behaviour I observe since I have it in production). If #880 fix the issue and also fix another one this is way better and you should definitely go for it !

If you don't have the time to investigate on the issue you sometimes get, could you tell me how to reproduce it ? Even if not deterministic ? I might be able to find some time to try to spot the error (if you cannot merge #880 right away).

Thanks again for your time

fzumstein:
yes, I've merged #881 instead as it seems a lot more robust. will try to release by tomorrow.

Colin-b:
Thanks a lot !

PR #873, opened 2018-05-14
Allow to retrieve Excel caller in arguments (is_excel_caller)

Colin-b:
This will allow to send results to Excel in an asynchronous way when using UDFs


njwhite:
FWIW you can get the Range of an argument with a custom converter (overriding reader instead of just read_value and pulling range out of the ConversionContext you get given)

fzumstein:
what about http://docs.xlwings.org/en/stable/udfs.html#the-vba-keyword

njwhite:
@fzumstein well, that’s just what I was looking for :)

PR #933, opened 2018-09-02
Use logger library over print(..)

njwhite:
It's easier to configure what messages are visible and where they go.

### DIFF
--- a/xlwings/server.py
+++ b/xlwings/server.py
@@ -6,6 +6,7 @@ try:
     import _win32sysloader
 except:
     raise Exception("Cannot import PyWin32. Are you sure it's installed?")
+import logging
 import sys
 import os
 # Hack to find pythoncom.dll - needed for some distribution/setups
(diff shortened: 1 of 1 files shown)


fzumstein:
How are you using this? The advantage of using `print` is that you can simply change the interpreter from `pythonw` to `python` and you get the messages in the console that pops up.

Colin-b:
The advantage of using logger is that your clients can decide where to redirect/save logs, filter them, change the formatting, etc....

Logging module provide everything print function provides, but print only provide a small subset of what logging provides.

fzumstein:
How do you configure the logging handler when `debug=False`?

Colin-b:
Via the python file called by xlwings. This one is setting the configuration

fzumstein:
yeah, it does work e.g. for file based logging, but I don't manage to get it to the console when using `python` as the interpreter. And I would like to keep that.

Colin-b:
Using a StreamHandler allows you to output to stdout. Is your question related to setting it as the default handler for xlwings users ?

fzumstein:
No, my question is how to get it to show in the command prompt that pops up with `python` as interpreter. `logging.basicConfig(level=logging.DEBUG)` works when in debug mode but not otherwise.

fzumstein:
replaced by #943

PR #939, opened 2018-09-25
Preserve docstring of UDFs in non windows environments

ludaavics:
When working in windows environment, the `@xw.func` decorator for UDFs [preserves the docstring](https://github.com/ZoomerAnalytics/xlwings/blob/master/xlwings/udfs.py#L119).

In non windows environment, since UDFs are not supported, the decorator is replaced by a [dummy function](https://github.com/ZoomerAnalytics/xlwings/blob/master/xlwings/__init__.py#L38).

However, the docstring may still be useful for generating automated documentation, for instance with [Sphinx](http://www.sphinx-doc.org/en/master/)'s `.. automodule::` directive.

### DIFF
--- a/xlwings/__init__.py
+++ b/xlwings/__init__.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from functools import wraps
 import sys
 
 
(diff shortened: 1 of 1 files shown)


fzumstein:
made the same change for sub and the dummy decorators were actually broken anyways, so I fixed that, too: https://github.com/ZoomerAnalytics/xlwings/commit/094610f3ecb5ba66306392970140a57bc4513f01

PR #942, opened 2018-10-02
raise exception in case of jagged arrays

fzumstein:


fzumstein:
I'd expect most users to work with numpy/pandas when using huge data. If it turns out to be an issue we can always roll back.

PR #978, opened 2018-11-04
added support for func(async='threading'), closes #473

### LINKED ISSUE #473: Support for asynchronous User-Defined Functions (Excel 2010)
Excel 2010 introduced asynchronous User-Defined Functions, which can be extremely useful for long-running tasks:

https://msdn.microsoft.com/en-us/library/office/ff796219%28v=office.14%29.aspx

I noticed that [@ericremoreynolds commented async UDFs were on your roadmap](https://github.com/ZoomerAnalytics/xlwings/issues/362#issuecomment-188130042), but I still wanted to open an issue to track this specific request and perhaps know if you have an ETA in mind.

I guess [the new async constructs in Python 3.5](https://docs.python.org/3/whatsnew/3.5.html#whatsnew-pep-492) would be a nice starting point, but maybe others can come up with an implementation which is compatible with [Legacy Python](http://carreau.github.io/posts/planning-an-early-death-for-python-2.html) as well.

cc @jmatiasindizen @skasi7

fzumstein:


fzumstein:
sure, give it a try and let me know if you see any issues

PR #995, opened 2018-11-22
fix when caller is None

fzumstein:


Colin-b:
In what case can a caller be none? VBA direct call? Also it seems there are some other places were caller and cache_key are used some lines after this :-)

fzumstein:
caller is None outside of Excel, e.g. Access VBA. The other occasions should not be hit then anyways as they are within dynamic or async parts.

PR #1016, opened 2018-12-20
Catch getmtime errors during module reload

Pluckerpluck:
On first loading of a module, this catches any errors generated using `getmtime()` and sets `filetime` to `None`. On subsequent reloads it checks `filetime` to see if it's `None`, and if it is we assume that the file cannot change and thus does not need to reload.

This does not fix `Import UDFs` in all cases (as that can break for other reasons), but it does allow the packaging of a module into a zip file for distribution, and not have xlwings immediately crash when it tries to run anything.

There should be no detrimental effect to current functionality.

### DIFF
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -212,29 +212,35 @@ class DelayedResizeDynamicArrayFormula(object):
 def get_udf_module(module_name):
     module_info = udf_modules.get(module_name, None)
     if module_info is not None:
-        mtime = os.path.getmtime(module_info['filename'])
         module = module_info['module']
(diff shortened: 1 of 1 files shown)


fzumstein:
I've added a change to include mybook.zip by default in the pythonpath: https://app.xltrail.com/#/workbook/github.com%2FZoomerAnalytics%2Fxlwings.git%2Fxlwings%2Faddin%2Fxlwings.xlam/diff/32b8d8dc020a1d302df4e9809143e6e33b46d33c/7b3425a4f365487b40098d5740d1df131a8110c3?public=true

PR #1024, opened 2019-01-29
Error handling when range has formulas which return #NA. 

shirisha95:
While accessing ws.range("A2:A100") if formulas can't be evaluated - all of them returns NA. we get an exception as we try and access -> c.value[0] in line 109 which results in this error

### DIFF
--- a/xlwings/conversion/standard.py
+++ b/xlwings/conversion/standard.py
@@ -104,7 +104,9 @@ class AdjustDimensionsStage(object):
         # the assumption is that value is 2-dimensional at this stage
 
         if self.ndim is None:
-            if len(c.value) == 1:
+            if len(c.value) == 0:
+                c.value = ""
+            elif len(c.value) == 1:
                 c.value = c.value[0][0] if len(c.value[0]) == 1 else c.value[0]
             elif len(c.value[0]) == 1:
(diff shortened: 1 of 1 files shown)


fzumstein:
Can you please attach a sample workbook to reproduce (or write a test)?

fzumstein:
@shirisha95 did you see my comment?

shirisha95:
Code for testing:
`import xlwings as xw
wb = xw.Book('test8.xlsx')
ws = wb.sheets['Summary']
print(ws.range("B2:B4").value)
wb.save()`

[test8.xlsx](https://github.com/ZoomerAnalytics/xlwings/files/2854619/test8.xlsx)

fzumstein:
I get the following output of your test function with the test code:
```
[-2146826246, -2146826246, -2146826246]
```

Which is logged as this bug (which should admittedly get fixed): https://github.com/ZoomerAnalytics/xlwings/issues/61

fzumstein:
I get the following output of your test function with the test code:
```
[-2146826246, -2146826246, -2146826246]
```

Which is logged as this bug (which should admittedly get fixed): https://github.com/ZoomerAnalytics/xlwings/issues/61

fzumstein:
replaced by #1060

PR #1051, opened 2019-03-15
Allow installation of xlwings on Linux even though it can't be used

SimonBiggs:
Hi :),

This library has been massively helpful for bridging the gap between Python code and Excel users. I massively appreciate it, thank you.

I am building a library called PyMedPhys (https://github.com/pymedphys/pymedphys). I develop this library on Windows and Linux, and run CI on Windows, Linux and Mac. Part of the testing framework is to run pylint within pytest. I understand that xlwings is not supported on Linux, however, when it is installed on Linux it does still enable rudimentary sanity checking via pylint.

I am very keen to be able to have xlwings installed on Linux even just so that it can be used by pylint.

Hence this pull request.

Cheers,
Simon

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,12 +25,8 @@ elif sys.platform.startswith('darwin'):
     install_requires = ['psutil >= 2.0.0', 'appscript >= 1.0.1']
     data_files = [(os.path.expanduser("~") + '/Library/Application Scripts/com.microsoft.Excel', ['xlwings/xlwings.applescript'])]
 else:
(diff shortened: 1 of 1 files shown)


fzumstein:
Can you please make this conditional on an environment variable (something like `INSTALL_ON_LINUX=1`)? There are already enough people today confused with the very clear error message, see: https://stackoverflow.com/questions/55121618/could-not-install-xlwings-in-my-ubuntu-machine

fzumstein:
replaced by #1052

SimonBiggs:
Thanks Felix :)

PR #1085, opened 2019-04-23
Create export_to_csv

Charles-Fox1234:
This function allows the user to skip the "save as csv, save as xlsm" process to create a csv file.  Instead, it can be done with this udf.


fzumstein:
While this is cool, I rather see this as a recipe in the docs (or as a GitHub gist), it shouldn't be part of a general purpose library. I would also add a docstring so that it's clear on how to use types.

PR #1094, opened 2019-04-30
Add an example for working with named ranges and sql

Charles-Fox1234:
This shows how xlwings makes reconciliation easier.


fzumstein:
this looks to complex to me for an example and I wouldn't want to use `.api` all over the place, i.e. would want to implement the missing methods first natively.

PR #1118, opened 2019-06-18
Fixing bug of not being able to use vba=application and *args

sbremer:
Bug: varargs does not work in combination with a "vba=" statement, for example to get the "Application" datastructure from excel.

Example:

@xw.arg('xl_app', vba='Application')
def vararg_func(xl_app, name, *args):
    pass

Fix:
Changed the Python code that generates the vba function wrappers.
Just setting all "argsArray" elements separately did not work for datastructures like "Application".
Workaround: Build an Array(...) first, then Redim into desired size and fill with varargs.

Convention: varargs "*args" must be last in enumeration of arguments in Python!

### DIFF
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -379,24 +379,15 @@ def generate_vba_wrapper(module_name, module, f):
                         vba.writeln('Application.Volatile')
 
                 if vararg != '':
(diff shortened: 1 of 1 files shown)


fzumstein:
thanks - I'll try to review soon!

PR #1120, opened 2019-06-19
does not match release dll files

inkenbrandt:
VBA error occurs despite repeated clean installs in conda. Replace "dev" with actual version number should resolve dll name mismatch.


fzumstein:
Hi Paul, you can't install directly from source. Instead, download the built packages from PyPI: https://pypi.org/search/?q=xlwings

Issue #322, opened 2015-12-09
cannot detect if called from excel


simonm3:
I have an logging handler that writes a log to excel (see attached). I include this in my logging configuration for a number of packages. It is intended to be silent if not needed; or to log output to excel if called from a worksheet by xlwings.

To detect if called from excel I do this:

        try:
            _ = Workbook.caller()
            self.calledFromExcel = True
        except:
            self.calledFromExcel = False
            return

The latest version of xlwings.bas changed the API for Workbook.caller() making the above fail silently. There should be a named exception here or some other way of detecting if called from excel.
[xllogging.txt](https://github.com/ZoomerAnalytics/xlwings/files/56668/xllogging.txt)

fzumstein:
Ummm are you sure that it's not just a mismatch between the VBA module version and the Python package version? We currently don't have a warning mechanism if they are out of sync.

simonm3:
Yes it is a mismatch which I have now corrected.

In this case the call to Workbook.caller() failed for a reason other than
"not called from excel". Therefore would be better to throw a named
exception to say it is not called from excel. In particular this is
important if there is potential for mismatch in versions between VBA module
and python package which is not checked.

fzumstein:
sure that's right, but even in your current version you should not have a naked except but catch the specific error which is raised (even if that one is not very descriptive at the moment)

simonm3:
If there was a specific error raised I would catch it; or if there is
another way to establish if it is called from Excel or not then I will use
it. Currently there is only a general exception thrown so that is what I
trap.

fzumstein:
sorry, yes, you're right, makes sense!

writetoprithvi:
Hi fzumstein, I am receiving the same error, when I call python module from excel. Can you help? Thanks

Issue #859, opened 2018-05-04
Propose a more specific Exception for invalid xw.Book.caller() calls.


Kamforka:
Hey guys,

The past few days I was doing some look around in the `xlwings` module and when I tried to set up a `try - except` block to catch an exception for a call to `Book.caller()` with the caller being unset I realized that the the exception is being kinda broad, so I guess a more specific one should be used instead.

Current behavior:

```python
import xlwings as xw
>>> xw.Book.caller()
Traceback (most recent call last):
  File "<pyshell#46>", line 1, in <module>
    xw.Book.caller()
  File "C:\...\Python\Python36\lib\site-packages\xlwings\main.py", line 554, in caller
    raise Exception('Workbook.caller() must not be called directly. Call through Excel or set a mock caller '
Exception: Workbook.caller() must not be called directly. Call through Excel or set a mock caller first with Book.set_mock_caller().
```

As for now a plain `Exception` is being raised for the invalid call.

My proposal would be to change it to something like: `CallerError` or `CallerException` or anything else that is meaningful here.

Regards,
Kamfor

fzumstein:
This is kind of a duplicate of https://github.com/ZoomerAnalytics/xlwings/issues/322, the issue is also logged on the roadmap, see: https://github.com/ZoomerAnalytics/xlwings/issues/324 so I am closing this.

Issue #324, opened 2015-12-10
Roadmap v1.0.0


fzumstein:
This is only an overview, the individual steps are/should be logged in separate issues (links will follow).  

- [ ] Mac Excel 2016 RunPython calls
- [ ] Accessors (nparray, dataframe, ...)
- [ ] Runtime add-in as alternative to VBA module incl. version check between VBA and Python
- [ ] Handle the potential EXCEL.EXE zombie processes when using global App/Workbook vars
- [ ] API refactoring
- [ ] xlwings Exceptions

a093130:
I try to access named ranges and named sheets from the xlwings python API.  No guarantee the workbook opened will be in the format I require.  So looking for exceptions.
example:
    try:
        wb = xw.Book(fname)
        sht = wb.sheets('GMAT')
        return sht.range('A1').expand().value
    except com_error:
        logging.error('Could not access range on workbook sheet GMAT.')        

I think xlwings is throwing an exception called com_error when the sheetname is bad, but python does not recognize it.  Example:

ohno = cfs.sheeterror(fname[0])
Traceback (most recent call last):

  File "<ipython-input-161-3818aba80da5>", line 1, in <module>
    ohno = cfs.sheeterror(fname[0])

  File "C:\Users\chelm\Documents\WorkingFolder\ProjectRoot\projects\Python3\GMAT-Automation\fromconfigsheet.py", line 74, in sheeterror
    sht = wb.sheets('bad')

  File "C:\Users\chelm\Anaconda3\Lib\site-packages\xlwings\main.py", line 2783, in __call__
    return Sheet(impl=self.impl(name_or_index))

  File "C:\Users\chelm\Anaconda3\Lib\site-packages\xlwings\_xlwindows.py", line 511, in __call__
    return Sheet(xl=self.xl(name_or_index))

  File "C:\Users\chelm\Anaconda3\Lib\site-packages\xlwings\_xlwindows.py", line 151, in __call__
    v = self._inner(*args, **kwargs)

  File "C:\Users\chelm\Anaconda3\Lib\site-packages\win32com\client\dynamic.py", line 197, in __call__
    return self._get_good_object_(self._oleobj_.Invoke(*allArgs),self._olerepr_.defaultDispatchName,None)

com_error: (-2147352567, 'Exception occurred.', (0, None, None, None, 0, -2147352565), None)


[...]

PR #1123, opened 2019-06-23
Implement custom exceptions for caller errors

zoquda:
Errors during during calls between python and an excel app would raise a
generic `Exception` error. Custom exceptions are implemented here
instead in order to enable users to perform more robust error catching.

A dedicated `exceptions` module and a base exception class are
introduced in order to facilitate implementation of further custom
exception classes in the future. The code organization is based on the
guidelines from Julien Danjou's "The definitive guide to Python
exceptions" (https://julien.danjou.info/python-exceptions-guide/)

Closes: #322
See also: #324, #859

### DIFF
--- a/xlwings/__init__.py
+++ b/xlwings/__init__.py
@@ -45,10 +45,6 @@ else:
 
 time_types = xlplatform.time_types
 
-# Errors
-class ShapeAlreadyExists(Exception):
-    pass
-
 # API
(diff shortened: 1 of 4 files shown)


fzumstein:
Great, thanks for pushing this. I would, however, like to use this to get rid of all generic Exceptions that we currently have in the code, possibly with wrapping exceptions of pywin32/appscript, as described in your blog post.

fzumstein:
This is on the roadmap, but I'd like to introduce it across the board, so closing this PR in the meantime.

Issue #1124, opened 2019-06-23
Add support for `pathlib.Path` file paths


zoquda:
#### Description

Currently the xlwings API does not seem to support file paths that are formatted as [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html)  object. The following code example demonstrates this:
```python
# OS: Windows 10
# xlwings: 0.15.8
# excel: 2013
# python: 3.7.3

from pathlib import Path
import xlwings as xw

# Create dummy excel file
book = xw.Book()
book.save('example_file.xlsx')
book.close()

# Attempt to connect to excel file using Pathlib path
file_path = Path('example_file.xlsx').resolve()
xw.Book(file_path)
```

The subsequent error traceback is:

```
AttributeError                            Traceback (most recent call last)
~\Desktop\example.py in <module>
      9 # Attempt to connect to excel file using Pathlib path
     10 file_path = Path('example_file.xlsx').resolve()
---> 11 xw.Book(file_path)

C:\Software\Anaconda3\lib\site-packages\xlwings\main.py in __init__(self, fullname, impl)
    474         if not impl:
    475             if fullname:
--> 476                 fullname = fullname.lower()
    477
    478                 candidates = []

AttributeError: 'WindowsPath' object has no attribute 'lower'
```

The addition of support for `pathlib.Path` file paths is hereby proposed as enhancement of xlwings.

#### Rationale

Use of `pathlib.Path` objects for file paths is attractive to users because (compared to string object file paths) they are usually easier to work with and more consistent across operating systems (see [Python 3's pathlib Module: Taming the File System](https://realpython.com/python-pathlib/) for more details). The `pathlib` module is readily available, since it is part of python's standard library since version 3.4.

PR #1126, opened 2019-06-25
Add support for path-like objects in API

zoquda:
This pull request adds support for [path-like objects](https://www.python.org/dev/peps/pep-0519/) (available in python 3.6 and higher) to API input arguments. It should now be possible for users to provide any object that is an instance of [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) (which includes [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html) objects) as path argument.

The changes made basically amount to ensuring that any path-like object is converted to a string object at the earliest opportunity.

Included as well are some new unit tests with `pathlib.Path` objects.

Closes: #1124

### DIFF
--- a/xlwings/main.py
+++ b/xlwings/main.py
@@ -464,7 +464,7 @@ class Book(object):
 
     Parameters
     ----------
-    fullname : str, default None
+    fullname : str or path-like object, default None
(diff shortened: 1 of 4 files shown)


zoquda:
May I ask what the assessment of this pull request is?

fzumstein:
sorry for the delay! I should have time soon to get to it!

PR #1145, opened 2019-09-04
added zh_hant_tw (traditional chinese) translation for installation guide

yuyueugene84:


fzumstein:
hey sorry for slow reaction! just one thing: shouldn't it be in the directory `docs/locale` instead of `docs/locales`?

yuyueugene84:
@fzumstein Fixed it, sorry for the late reply!

PR #1177, opened 2019-11-11
Use formula cache to prevent disappearing legacy dynamic array

### LINKED ISSUE #1164: UDF with expand=table disappears on recalculation
#### OS (e.g. Windows 10 or macOS Sierra)

Windows 10 Enterprise

#### Versions of xlwings, Excel and Python (e.g. 0.11.8, Office 365, Python 3.7)

xlwings 0.15.10, Excel Professional Plus 2016, Python 3.7

#### Describe your issue (incl. Traceback!)

Thanks for including the expand functionality in the library, it seems very useful if I can get it working smoothly.

I had noticed as I was building a workbook using UDFs with `expand='table'` that sometimes my UDFs would disappear entirely upon recalculation. I had to do a lot of experimenting to narrow down exactly when this occurs.

This happens for UDFs with `expand='table'` for both `@xw.ret` and `@xw.arg` if they are referencing another UDF which has `expand='table'` for `@xw.ret`. It also only occurs if this is the last function which was defined referencing the cells of the table. 

E.g. the setup looks something like this:
Plain Cells -> Reference plain cells by UDF with `expand='table'` for both `@xw.ret` and `@xw.arg` -> Reference first UDF output by another UDF with  `expand='table'` for `@xw.ret`.

If any of these conditions are not met, the problem will not occur. Calculating any function, UDF or otherwise (=SUM works), on the full cell range of the table after defining the UDF will make the UDF call not disappear. If the same UDF is called a second time, the second occurrence will disappear, but the first will not.

If `expand='table'` is excluded for either  `@xw.ret` or `@xw.arg` on the first UDF in the chain, the problem will not occur. If `expand='table'` is excluded for  `@xw.ret` on the second UDF, the problem will not occur.

A temporary workaround on the test project is to call my UDF first in the spot where I want it, then call it in an empty spot that I don't want it. Upon recalculate, the one in the empty spot disappears, but the first one will always persist from that point on. However there may be still some conditions where this occurs that I don't understand, as I have tried this fix in my full workbook, and it works initially, but after recalculating many times, some functions will still disappear. In this workbook I have a chain of 4 UDFs using expand.

#### Include a minimal code sample to reproduce the issue (and attach a sample workbook if required!)

While this will happen for any UDF matching the conditions above, I have included the most simple version possible here.

```python
@xw.func
@xw.arg('data', expand='table')
@xw.ret(expand='table', index=False)
def print_table(data):
    return data
```

This sample project shows the issue. Open the workbook, and recalculate with `CTRL + ALT + F9`, and the second UDF disappears on my machine. 

Try opening the workbook again, if instead, you call `print_table` a second time referencing the first UDF, then only this newly created UDF will disappear on recalculation, and not ever the original second UDF again.

Try again, if you call `=SUM(E4:E6)`, then recalculate, the second UDF will not disappear. You can even delete the `=SUM` call then, and recalculate as many times as you want and the second UDF will not disappear.

[disappearing.zip](https://github.com/xlwings/xlwings/files/3735804/disappearing.zip)

I appreciate any help you could give. I could also try to fix this if you point me in the right direction.

nickderobertis:
This PR fixes #1164 in my testing.

It seems that the issue was that the resize operation is getting called multiple times, and in the test workbook attached in #1164, `DelayedResizeDynamicArrayFormula` would be able to get the formula on the first resize operation, but thereafter it was getting the formula as `None` when pulling from `self.caller.FormulaArray`.

I fixed this by implementing the formula caching approach from `xlwings-1010-improve-async`, though with a slight modification and a different spot. It is still in `call_udf` but now under the `if is_dynamic_array` block. Then `DelayedResizeDynamicArrayFormula` retrieves from cache if it gets `None` for the formula. So now we have function cache in addition to the existing result cache.

As described in `xlwings-1010-improve-async`, there is sometimes a `com_error` while getting the formula from caller during dynamic array resizing. I added error handling for if somehow we hit the error before the resizing, as then there would be no formula in the cache.

Then I went to remove the formula from the cache. I tried three approaches for this, settling on the third. The first was simply deleting the formula cache where we are deleting the result cache, right before the end of `call_udf`. Then I tried creating a function which does this and passing it to `add_idle_task` in the same spot. Both of these approaches still reach the exception I added for not being able to access `caller.FormulaArray` and also not having the formula in the cache (it is clearing the cache too early). 

The third approach for removing the formula cache, and the one currently used, builds on top of the idle task approach, but the function to delete also queries the idle queue to check if any tasks are still using the cache key. If any are, then it adds itself to the end of the queue for later deletion. If not, then it will move forward with deletion. 

I noticed in my debugging of the cache that the result cache was already not being handled properly, the result cache keys (which I didn't touch) were building up for various different results throughout the multiple calculations of the dynamic array. So I tried generalizing this safe cache delete and applying it to the existing result cache as well. For some reason, this caused `xlwings` to hang, so I reversed this change, now applying only to the formula cache. So now the state of the cache is as it was before this PR, cache keys build up for result cache, but there are no leftover formula cache keys (before these didn't exist).

I am open to suggestions to remove the leftover result cache as well, though this is going beyond the scope of fixing #1164 

I'm not sure if there are any side effects of this approach, or if I have covered all the cases. But it works in my testing and passes `nosetests`. I would be happy to make additional changes if necessary.

### DIFF
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -373,7 +373,13 @@ def call_udf(module_name, func_name, args, this_workbook=None, caller=None):
             except pywintypes.com_error:
                 # Expected to hit here during dynamic array resizing, but if cache was never set then will be an issue
                 if formula_cache_key not in cache:
-                    raise Exception(f'not able to get formula for {func_name} from caller or cache')
+                    # Sometimes still hitting here due to FormulaArray not being available, yet .formula is still available
+                    # .formula has repeated formulas in a tuple of tuples, one for each output cell. Extract the top left
+                    try:
+                        formula_array = caller.formula[0][0]
+                        cache[formula_cache_key] = formula_array
+                    except pywintypes.com_error:
+                        raise Exception(f'not able to get formula for {func_name} from caller or cache')
             cache_key = get_cache_key(func, args, caller)
             cached_value = cache.get(cache_key)
             if cached_value is not None:
@@ -404,8 +410,8 @@ def call_udf(module_name, func_name, args, this_workbook=None, caller=None):
         else:
             del cache[cache_key]
 
-        del_formula_cache = partial(safe_delete_from_cache, cache, formula_cache_key, 'formula_cache_key')
-        add_idle_task(del_formula_cache)
+            del_formula_cache = partial(safe_delete_from_cache, cache, formula_cache_key, 'formula_cache_key')
+            add_idle_task(del_formula_cache)
 
     return xl_result
 
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -358,6 +358,24 @@ def call_udf(module_name, func_name, args, this_workbook=None, caller=None):
     if this_workbook:
         xlplatform.BOOK_CALLER = Dispatch(this_workbook)
 
+    # Put dynamic array formula in cache as sometimes it cannot be retrieved from Excel during the calculation
+    if is_dynamic_array:
+        formula_cache_key = get_formula_cache_key(func, caller)
+        try:
+            # Can't get array_formula during dynamic array resizing
+            formula_array = caller.FormulaArray
+            cache[formula_cache_key] = formula_array
+        except pywintypes.com_error:
+            # Expected to hit here during dynamic array resizing, but if cache was never set then will be an issue
+            if formula_cache_key not in cache:
+                # Sometimes still hitting here due to FormulaArray not being available, yet .formula is still available
+                # .formula has repeated formulas in a tuple of tuples, one for each output cell. Extract the top left
+                try:
+                    formula_array = get_array_formula_from_formula(caller)
+                    cache[formula_cache_key] = formula_array
+                except pywintypes.com_error:
+                    raise Exception(f'not able to get formula for {func_name} from caller or cache')
+
     if func_info['async_mode'] and func_info['async_mode'] == 'threading':
         cache_key = get_cache_key(func, args, caller)
         cached_value = cache.get(cache_key)
@@ -380,21 +398,6 @@ def call_udf(module_name, func_name, args, this_workbook=None, caller=None):
             return [["#N/A waiting..." * xw_caller.columns.count] * xw_caller.rows.count]
     else:
         if is_dynamic_array:
-            formula_cache_key = get_formula_cache_key(func, caller)
-            try:
-                # Can't get array_formula during dynamic array resizing
-                formula_array = caller.FormulaArray
-                cache[formula_cache_key] = formula_array
-            except pywintypes.com_error:
-                # Expected to hit here during dynamic array resizing, but if cache was never set then will be an issue
-                if formula_cache_key not in cache:
-                    # Sometimes still hitting here due to FormulaArray not being available, yet .formula is still available
-                    # .formula has repeated formulas in a tuple of tuples, one for each output cell. Extract the top left
-                    try:
-                        formula_array = get_array_formula_from_formula(caller)
-                        cache[formula_cache_key] = formula_array
-                    except pywintypes.com_error:
-                        raise Exception(f'not able to get formula for {func_name} from caller or cache')
             cache_key = get_cache_key(func, args, caller)
             cached_value = cache.get(cache_key)
             if cached_value is not None:
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -206,22 +206,14 @@ class DelayedResizeDynamicArrayFormula(object):
         self.caller = caller
         self.needs_clearing = needs_clearing
         self.formula_cache_key = formula_cache_key
-        self.uuid = uuid.uuid4()
-        print(f'Creating task {self.uuid}')
 
     def __call__(self, *args, **kwargs):
-        print(f'Calling task {self.uuid}')
         formula = self.caller.FormulaArray
         if formula is None:
             # Could not read formula properly, get from cache
-            print('Getting formula from cache')
             formula = cache.get(self.formula_cache_key)
-        else:
-            print('Got formula from caller')
         if self.needs_clearing:
-            print(f'Clearing formula {formula}')
             self.caller.ClearContents()
-        print(f'Setting formula to {formula}')
         self.target_range.api.FormulaArray = formula
 
 
@@ -325,7 +317,9 @@ def call_udf(module_name, func_name, args, this_workbook=None, caller=None):
                 formula_array = caller.FormulaArray
                 cache[formula_cache_key] = formula_array
             except pywintypes.com_error:
-                print('could not cache formula array')
+                # Expected to hit here during dynamic array resizing, but if cache was never set then will be an issue
+                if formula_cache_key not in cache:
+                    raise Exception(f'not able to get formula for {func_name} from caller or cache')
             cache_key = get_cache_key(func, args, caller)
             cached_value = cache.get(cache_key)
             if cached_value is not None:
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -312,6 +312,21 @@ def safe_delete_from_cache(cache, key, task_key_attr='formula_cache_key'):
         pass
 
 
+def get_array_formula_from_formula(caller):
+    """
+    Extract array for formula from caller.formula or caller.Formula. Used when caller.FormulaArray is giving com_error
+    """
+    try:
+        # e.g. caller: <COMObject Range>
+        formula_tup = caller.formula
+    except AttributeError:
+        # e.g. caller: <win32com.gen_py.Microsoft Excel 16.0 Object Library.Range instance at 0x695150436648>
+        formula_tup = caller.Formula
+
+    # formula is a tuple of tuples for the number of cells, but all formulas are the same so take the top left.
+    return formula_tup[0][0]
+
+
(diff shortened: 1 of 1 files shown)


whoopnip:
This was all working fine for me as of `58cedc1fd815973009091813c4da41697b126e6b` on the test workbook in #1164 but when I went to go try it on my full workbook, with several interlinked dynamic array UDFs, upon many recalculations it was sometimes failing to get the formula from caller.FormulaArray even though there was no formula in the cache. I added the following two commits to fix this. When I put a breakpoint where this was happening, caller.FormulaArray gave:

`pywintypes.com_error: (-2147352567, 'Exception occurred.', (0, 'Microsoft Excel', 'Unable to get the FormulaArray property of the Range class', 'xlmain11.chm', 0, -2146827284), None)
`

But `caller.formula` had the expected formula, though in a tuple of tuples format for each cell.

Therefore I have added extracting `formula_array = caller.formula[0][0]` when `caller.FormulaArray` raises a `com_error` and the formula is not already in cache. This is working well for me in my full workbook.

whoopnip:
Also for some reason, when I use this through my `xlwings` development setup which was set up using `python setup.py develop`, when it cannot find `caller.FormulaArray`, I am getting the caller as `<COMObject Range>`, while when I install it normally, then in the same situation the caller comes as `<win32com.gen_py.Microsoft Excel 16.0 Object Library.Range instance at 0x695150436648>`.

The `COMObject Range` has a lower case `.formula` attribute, while the  `win32com.gen_py.Microsoft Excel 16.0 Object Library.Range` has an upper case `.Formula` attribute. I added code in `e97f0a090ead9e4782a254e60e8ed0341cc393a7` to handle looking up both `.formula` and `.Formula`, whichever is available. Now this is working fine when installed normally as well as in development mode.

fzumstein:
great, let me know when you think this is ready for me to test.

whoopnip:
I think it's ready now. The test workbook from #1164 works fine now, and I've been running Monte Carlo simulations on my full workbook with multiple interlinked dynamic UDFs using this code, so I've recalculated it nearly 100,000 times without the functions disappearing. It used to be that a few recalculations would cause one of the UDFs to disappear.

fzumstein:
One issue I've found is when you open `tests/udfs/async_tests.xlsm` you will see that they break with

```
"UnboundLocalError: local variable 'formula_cache_key' referenced before assignment
    del_formula_cache = partial(safe_delete_from_cache, cache, formula_cache_key, 'formula_cache_key')
  File ""c:\users\felix\dev\xlwings\xlwings\udfs.py"", line 428, in call_udf
    res = call_udf(script, fname, args, this_workbook, FromVariant(caller))
  File ""c:\users\felix\dev\xlwings\xlwings\server.py"", line 198, in CallUDF
    return func(*args)
  File ""C:\Users\felix\Miniconda\lib\site-packages\win32com\server\policy.py"", line 586, in _invokeex_
    return S_OK, -1, self._invokeex_(dispid, lcid, wFlags, args, None, None)
  File ""C:\Users\felix\Miniconda\lib\site-packages\win32com\server\policy.py"", line 283, in _invoke_
    return self._invoke_(dispid, lcid, wFlags, args)
  File ""C:\Users\felix\Miniconda\lib\site-packages\win32com\server\policy.py"", line 278, in _Invoke_"
```

whoopnip:
I see, thanks for testing this. Are the async tests not run with `nosetests`? Because those worked fine for me. I will make sure to fix this and also check `tests/udfs/async_tests.xlsm` this time. 

I think the fix will just be moving my logic which sets up formula cache to before the block which is checking for async rather than within the else clause. It would then be in its own `if is_dynamic_array` block [(at line 360 in my fork](https://github.com/whoopnip/xlwings/blob/xlwings-1164-disappearing-dynamic-array/xlwings/udfs.py#L360)). @fzumstein do you know offhand if any of `caller.FormulaArray`, `caller.Formula`, or `caller.formula` will be available for an async UDF the first time it is called? If so then I think this approach will work fine.

fzumstein:
yeah, anything in the `tests/udfs` directory currently doesn't run as part of the test suite but needs to be tested manually. I am (off and on) working on getting a refactored test setup up via CI, so can fix this then. `caller.Formula` should be fine I think.

whoopnip:
OK so it still wasn't exactly clear how to run the UDF tests, my general process to run the UDF tests was to run each workbook manually in debug mode in my development virtualenv. After opening the workbook, I ran `CalculateFull()` to get all the UDFs to evaluate, checking for any errors in the debug terminal. After no errors, I would run `CalculateFull()` 50 times in a row to check for the quick calculation issue that was previously making the UDFs disappear, again checking for errors.

Then `udf_tests.xlsm` was a bit different, it seemed to have actual test results in the workbook, so there I scrolled through and checked that the tests were showing passing as `TRUE`. 

I tested the three workbooks in `tests/udfs` and then ran `nosetests` again.

Originally when I ran `async_tests`, I got the same error that you did. After moving the block to set the formula cache as I suggested, now I do not get any errors in the terminal, and all the tests seem to pass.

I think this should be good to go again for your testing, but let me know if I missed anything in the testing and I can do that.

fzumstein:
That fixes the error yes, however... I only now realised that you are using `@xw.arg('data', expand='table')` which isn't something that we want to support with UDFs. If you need a dynamic input, then you can use Excel tables, for example, and refer to the name of the table as input...so the fix in eyes should be to raise an error if `expand` is used with `xw.arg`.

whoopnip:
Wait, so you're basically saying this functionality is not needed? Or why are we not trying to support it?

My use case for this is I have one dynamic UDF outputting a changing number of rows and columns, then a second UDF needing to accept that first UDF's output as an input. This was straightforward when I could reference its top-left cell and use `@xw.arg('data', expand='table')`. 

So you are suggesting that whenever a user has this situation, there needs to be some additional code added to convert the output range of the first UDF to an Excel table and name that table, so that the second UDF can reference it? How would that code look?

It seems it would be much easier on the user in these kinds of situations to support `xw.arg` with `expand='table'`.

fzumstein:
You're right, excel tables won't work for that neither. 
Rather than building that hack into the library, you *might* get lucky with something along these lines:

```
@xw.func
@xw.arg("x", xw.Range)
@xw.ret(expand='table')
def myfunc(x):
    return x.expand().value
```

Sorry for not having spotted the `expand` in the args decorator from the beginning.

whoopnip:
So I tried your code sample, and it did work, but it produced exactly the same results as using `xw.arg` with `expand`. Without this branch, it runs into the same problem of disappearing after a few recalculations. With this branch, it does not disappear. 

Therefore regardless I think this PR is still needed. 

Further, since using `xw.arg` with `expand` is already working and produces equivalent results with what you suggest as the "correct" way to do this, I would request to not raise an error upon using `xw.arg` with `expand`. There may be others using this already, and it would be a pain for them to go and update their code just to have it function exactly the same. Further, I think it is more consistent with the standard UDF syntax in general to do this via decorator.

fzumstein:
Ok I've looked again at this branch. When I run your sample code from #1164 with an additional print line I see this: 

function:
```
@xw.func
@xw.arg('data', expand='table')
@xw.arg('xl', vba='Application')
@xw.ret(expand='table', index=False)
def print_table(data, xl):
    print(f'called func from {xl.Caller.Address}')
    return data
```

log:

```
xlwings server running, clsid={261ADCE2-36A5-4F75-A15B-911E89648CCD}
called func from $H$4:$I$6
called func from $E$4:$F$6
Retrying TaskQueue '<xlwings.udfs.DelayedResizeDynamicArrayFormula object at 0x00000282722B9708>'.
Retrying TaskQueue '<xlwings.udfs.DelayedResizeDynamicArrayFormula object at 0x00000282722B9708>'.
called func from $H$4:$I$6
Retrying TaskQueue '<xlwings.udfs.DelayedResizeDynamicArrayFormula object at 0x00000282722B9708>'.
called func from $H$4:$I$6
Retrying TaskQueue '<xlwings.udfs.DelayedResizeDynamicArrayFormula object at 0x00000282722B9708>'.
called func from $H$4:$I$6
Retrying TaskQueue '<xlwings.udfs.DelayedResizeDynamicArrayFormula object at 0x00000282722B9708>'.
called func from $H$4:$I$6
Retrying TaskQueue '<xlwings.udfs.DelayedResizeDynamicArrayFormula object at 0x00000282722B9708>'.
called func from $H$4:$I$6
Retrying TaskQueue '<xlwings.udfs.DelayedResizeDynamicArrayFormula object at 0x00000282722B9708>'.
called func from $H$4:$I$6
Retrying TaskQueue '<xlwings.udfs.DelayedResizeDynamicArrayFormula object at 0x00000282722B9708>'.
called func from $H$4:$I$6
called func from $H$4
called func from $H$4:$I$6
called func from $H$4:$I$6
called func from $H$4:$I$6
called func from $H$4:$I$6
called func from $H$4:$I$6
```

Is that what you see on your end, too?

whoopnip:
Hey, was busy wrapping up the semester but just got a chance to look at this. So I do indeed see the same output, though it varies a bit depending on exactly how I recalculate. 

But I checked this on the master branch, and got the exact same behavior, so this does not seem to be caused by this PR. There still seems to be some weirdness going on with the calculations, which hopefully could be fixed in a future PR.

fzumstein:
replaced by #1318

PR #1179, opened 2019-11-13
increased com server timeout to 1200

### LINKED ISSUE #1168: Change the hard coded UDF server startup timeout to be configurable
There is a hard coded 1 minute timeout of starting python server process: 
https://github.com/xlwings/xlwings/blob/14d0e5663b50a5a3767e06c0e75f4273ab808584/src/xlwingsdll/config.cpp#L479-L503

I have a python installation on a network drive. Sometimes, it can take more than 1 minute to start the python process due to network slowness. It would be nice if the timeout can be configured like other parameters, so that the timeout can be increased for use case like mine which expects long start up time.

Thanks.

p.s. This is a great project. Thank you so much for all your hard work.

fzumstein:
Closes #1168

### DIFF
--- a/src/xlwingsdll/config.cpp
+++ b/src/xlwingsdll/config.cpp
@@ -476,8 +476,8 @@ void Config::ActivateRPCServer()
 				throw formatted_exception() << "AssignProcessToJobObject failed: " << GetLastErrorMessage();
 		}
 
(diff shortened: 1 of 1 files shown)


crayxt:
The comment line above changed line should be updated as well.

Issue #1149, opened 2019-09-13
test_restapi.py not compatible with Python < 3.6


jonathonw:
[test_restapi.py](https://github.com/xlwings/xlwings/blob/master/xlwings/tests/restapi/test_restapi.py#L9) makes extensive use of literal string interpolation (`f''` strings), which was first introduced in Python 3.6 and is incompatible with Python 2.7 and Python 3 prior to 3.6.

fzumstein:
true that, but not very critical since it's only the tests. I am currently trying to setup CI for the tests, so might fix it then.

fzumstein:
in the meantime, 3.5 is EOL and xlwings has dropped support for <3.6

PR #1181, opened 2019-11-14
suggest to use f-string

rogererens:
f-strings are in python since version 3.6

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -112,7 +112,7 @@ Writing a UDF in Python is as easy as:
 
     @xw.func
     def hello(name):
-        return 'Hello {0}'.format(name)
+        return f'Hello {name}'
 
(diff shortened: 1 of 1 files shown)


fzumstein:
our users use all kinds of different versions, so we try to be compatible with all of them. Also trying to prevent this kind of issues: https://github.com/xlwings/xlwings/issues/1149

Colin-b:
I know this is closed but is it planned to drop support for non supported python versions at some point ?

fzumstein:
We will drop 2.7 support in all new versions from 2020 in the open-source library according to https://python3statement.org. In return, we will introduce the xlwings PRO package that continues to support 2.7 - even for new versions, see: https://www.xlwings.org/pricing
No plans for dropping support of older Python 3 versions for now.

PR #1240, opened 2020-01-16
Handle all different possibilities for np.nan

### LINKED ISSUE #1116: np.float64('nan') is not handled correctly
Excel prints a 65535 instead (as in #15).

This is because of the following lines in prepare_xl_data_element (https://github.com/xlwings/xlwings/blob/171b10743224b60c85343d7df2fadf9be01e67fc/xlwings/_xlwindows.py#L1040)
```python
elif np and isinstance(x, np.generic):
        return float(x)
```
In that case, x is converted to a Python builtin float('nan'), but not to "".

I will try to produce a PR to fix this next week.

rxxg:
Fixes #1116, handle more nan types

### DIFF
--- a/xlwings/_xlmac.py
+++ b/xlwings/_xlmac.py
@@ -1189,7 +1189,7 @@ def clean_value_data(data, datetime_builder, empty_as, number_builder):
 def prepare_xl_data_element(x):
     if x is None:
         return ""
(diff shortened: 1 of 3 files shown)


fzumstein:
thank you!

PR #1255, opened 2020-02-14
make refers_to_range deal with quotes

kushal-kumaran:
The address produced with external=True includes quotes around the workbook/sheet pair in the form `'[book.xlsm]sheet'!$A$1`. The current regex results in sheet name that includes the trailing quote, which results in reference to a non-existent sheet.

This patch excludes the trailing quote from the group that is expected to match the sheet name.


fzumstein:
replaced by #1256: you were missing the fact that quotes are only added if the sheet contains e.g. a space. Also, unit tests were missing.

PR #1280, opened 2020-03-20
Handle Range.Formula2

### LINKED ISSUE #1277: Allow to use Threading and Dynamic Array Formulas
Hi,

For now dynamic array formulas are prefixed with the [implicit intersection operator (@)](https://support.office.com/en-us/article/implicit-intersection-operator-ce3be07b-0101-4450-a24e-c1c999be2b34) when using threading on generated user defined functions.
Meaning that a single value is shown instead of the expected array.

```python
import xlwings as xw

@xw.func(async_mode='threading')
def many_values():
    return ["test", "test2"]
```
image

The issue comes from the fact that [Range.Formula] is used when updating the formula to trigger a new call retrieving the cached result: https://github.com/xlwings/xlwings/blob/7af17370cef4bb887ec41bd37d5df72dd3b8a4c5/xlwings/udfs.py#L39

Leading to the underlying use of Range.Formula as in the following:
https://github.com/xlwings/xlwings/blob/7af17370cef4bb887ec41bd37d5df72dd3b8a4c5/xlwings/_xlwindows.py#L743

Switching to [Range.Formula2] fixes the issue.
image

I will gladly provide a pull request but I was wondering about the design approach for this fix, so first I would like to check with you if the following can be merged:

* Exposing `formula2` (as a property, the same way it is done for `formula`) inside [`xlwings._xlwindows.Range`](https://github.com/xlwings/xlwings/blob/7af17370cef4bb887ec41bd37d5df72dd3b8a4c5/xlwings/_xlwindows.py#L667) and [`xlwings._xlmac.Range`](https://github.com/xlwings/xlwings/blob/7af17370cef4bb887ec41bd37d5df72dd3b8a4c5/xlwings/_xlmac.py#L441) classes.
* In [`xlwings.udfs.AsyncThread.run`](https://github.com/xlwings/xlwings/blob/7af17370cef4bb887ec41bd37d5df72dd3b8a4c5/xlwings/udfs.py#L32), try to set `formula2` with the previous `formula2`, if it fails, fallback to previous behavior (setting `formula` with the previous `formula`).

However I would point out that I don't have access to a Mac then I am unable to test the behavior on Mac, that's why I am proposing to use a try/except block on the operation, as I am unaware of the result.

Also, I don't have access to a version of Microsoft Excel that does not have [Range.Formula2], so I am unable to test the behavior when [Range.Formula2] is not available, that's why I am proposing to use a try/except block on the operation, as I am unaware of the result.

According to my tests, [Range.Formula2] works fine with arrays and single value return types.

[Range.Formula]: https://docs.microsoft.com/en-us/office/vba/api/excel.range.formula
[Range.Formula2]: https://docs.microsoft.com/en-us/office/vba/api/excel.range.formula2

Colin-b:
fixes #1277 at the same time

### DIFF
--- a/tests/test_range.py
+++ b/tests/test_range.py
@@ -161,6 +161,10 @@ class TestRangeAttributes(TestBase):
         self.wb1.sheets[0].range('A1').formula = '=SUM(A2:A10)'
         self.assertEqual(self.wb1.sheets[0].range('A1').formula, '=SUM(A2:A10)')
 
+    def test_formula2(self):
+        self.wb1.sheets[0].range('A1').formula2 = '=UNIQUE(A2:A10)'
+        self.assertEqual(self.wb1.sheets[0].range('A1').formula2, '=UNIQUE(A2:A10)')
+
     def test_formula_array(self):
         self.wb1.sheets[0].range('A1').value = [[1, 4], [2, 5], [3, 6]]
         self.wb1.sheets[0].range('D1').formula_array = '=SUM(A1:A3*B1:B3)'
--- a/xlwings/_xlmac.py
+++ b/xlwings/_xlmac.py
@@ -525,6 +525,16 @@ class Range:
             self.xl.formula.set(value)
 
     @property
+    def formula2(self):
+        if self.xl is not None:
+            return self.xl.formula2.get()
+
+    @formula2.setter
+    def formula2(self, value):
+        if self.xl is not None:
+            self.xl.formula2.set(value)
+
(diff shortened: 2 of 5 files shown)


Colin-b:
Hi, did you had some time to look into this? Is there something more you would like me to add?

fzumstein:
sorry, this looks good, I'll try to merge and test this week!

fzumstein:
replaced by #1293

PR #1285, opened 2020-04-01
Add support to create xlsx with wps office application

angela-1:
When I use xlwings to process xlsx files. I got an error because I did not installed Microsoft Office Excel.

I am an WPS user. It is a softwae like Excel.I have made a modification in source code, and looks like it works fine on my tasks. Maybe you could have a look.

### DIFF
--- a/xlwings/_xlwindows.py
+++ b/xlwings/_xlwindows.py
@@ -292,7 +292,13 @@ class App:
             warn('spec is ignored on Windows.')
         if xl is None:
             # new instance
-            self._xl = COMRetryObjectWrapper(DispatchEx('Excel.Application'))
+
+            try:
+                self._xl = COMRetryObjectWrapper(DispatchEx('Excel.Application'))
+            except pywintypes.com_error:
+                # for system with WPS Office installed
+                self._xl = COMRetryObjectWrapper(DispatchEx('KET.Application'))
+
             if add_book:
                 self._xl.Workbooks.Add()
             self._hwnd = None


fzumstein:
Hi angela, have you tried to run the whole test suite, see: https://github.com/xlwings/xlwings/blob/master/DEVELOPER_GUIDE.md#tests

angela-1:
> Hi angela, have you tried to run the whole test suite, see: 

https://github.com/xlwings/xlwings/blob/master/DEVELOPER_GUIDE.md#tests

My environment is Windows 7, with WPS 2019 installed.


I did a test with `nosetests`, and I got the output [test.log](https://github.com/xlwings/xlwings/files/4425491/test.log) with 1 error and 4 failure.

I don't know why the error occured. But the failures, I want to explain:

> FAIL: test_integers: Covers GH 227
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "E:\home\angela\repo\xlwings\tests\test_conversion.py", line 48, in test_integers
>     self.assertEqual(self.wb1.sheets[0].range('A101').value, 10000000000000000000)
> AssertionError: None != 10000000000000000000

I tried to decrese one zero and it passed. 

> =======================================================
> FAIL: test_column_width (tests.test_range.TestRangeAttributes)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "E:\home\angela\repo\xlwings\tests\test_range.py", line 178, in test_column_width
>     self.assertEqual(None, result)
> AssertionError: None != 10.0
> 
> =====================================================
> FAIL: test_row_height (tests.test_range.TestRangeAttributes)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "E:\home\angela\repo\xlwings\tests\test_range.py", line 191, in test_row_
> height
>     self.assertEqual(None, result)
> AssertionError: None != 15.0
> 
> ====================================================
> FAIL: test_activate (tests.test_sheet.TestSheet)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "E:\home\angela\repo\xlwings\tests\test_sheet.py", line 91, in test_activate
>     self.assertEqual(xw.apps.keys()[0], self.app1)
> AssertionError: 7344 != <Excel App 7344>
> 
> ----------------------------------------------------------------------> 

These 3 failures, you can see that they got the right value.

After all, these tests was wrote for Microsoft Office Excel on Windows,
WPS is a another software and their implementation are different. 

```
        self.wb1.sheets[0].range('A1:B2').value = 'ensure cells are used'
        self.wb1.sheets[0].range('B2').column_width = 20.0
        result = self.wb1.sheets[0].range('A1:B2').column_width
        if sys.platform.startswith('win'):
            self.assertEqual(None, result)
        else:
            self.assertEqual(kw.missing_value, result)
```

The test code want to asert the result is None, but WPS gets right value.
 
In a word, I think this pr doesn't make errors to current code and add a support to WPS 
for users that do not use Microsoft Office Excel.

Whatever, many thanks for the help of xlwings.

fzumstein:
While I agree that the level of compatibility is impressive, it's not going to be 100%. I see for example that thing like `xw.books` don't work.

I am not going to open pandora's box here and seemingly add support for it, so I am closing this PR. But thanks for pointing this out and I think you should be able to manage to hack the source for this to work.

PR #1356, opened 2020-07-15
update code

efreeway:
update code for further translation


efreeway:
update for further translation

PR #1381, opened 2020-08-10
Added UDF import fallback to prevent unreproducable Automation error 440

fzumstein:
…error 440


Colin-b:
Won't this change make it impossible for other application to have their own version of xlwings ? I mean, I personally use another xlam with its own xlwings_udfs. This change will not allow that anymore, as it will require the xlam to be named as xlwings. Or am I mistaken ?

fzumstein:
better now?

Colin-b:
It's better, but I don't know how it will behave if the user have xlwings addin installed, and another xlam "embedding" xlwings ?  Not so sure how it would behave right now so I guess it's ok :-)

fzumstein:
shoud be good now

PR #1395, opened 2020-08-25
Fix COMRetryObjectWrapper

njwhite:
Currently it doesn't work with pywin32 COM generated dynamic dispatch, as the generated classes are subclasses of `CoClassBaseClass` and the code just tests for `is CDispatch` (generic dynamic dispatch). I've widened the check to DispatchBaseClass to cover this too.

### DIFF
--- a/xlwings/_xlwindows.py
+++ b/xlwings/_xlwindows.py
@@ -22,7 +22,7 @@ import types
 from ctypes import oledll, PyDLL, py_object, byref, POINTER, windll
 
 import pythoncom
-from win32com.client import Dispatch, CDispatch, DispatchEx
+from win32com.client import Dispatch, CoClassBaseClass, DispatchEx, DispatchBaseClass
 import win32timezone
 import win32gui
 import win32process
@@ -64,10 +64,9 @@ class COMRetryMethodWrapper:
         while True:
             try:
                 v = self.__method(*args, **kwargs)
-                t = type(v)
-                if t is CDispatch:
+                if isinstance(v, (CoClassBaseClass, DispatchBaseClass)):
                     return COMRetryObjectWrapper(v)
-                elif t is types.MethodType:
+                elif type(v) is types.MethodType:
                     return COMRetryMethodWrapper(v)
                 else:
                     return v
@@ -118,10 +117,9 @@ class COMRetryObjectWrapper:
         while True:
             try:
                 v = getattr(self._inner, item)
-                t = type(v)
-                if t is CDispatch:
+                if isinstance(v, (CoClassBaseClass, DispatchBaseClass)):
                     return COMRetryObjectWrapper(v)
-                elif t is types.MethodType:
+                elif type(v) is types.MethodType:
                     return COMRetryMethodWrapper(v)
                 else:
                     return v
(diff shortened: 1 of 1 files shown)


njwhite:
nb: fixes #1395

fzumstein:
This gives me the following error:

Edit a cell while running this code:

```
import xlwings as xw
xw.sheets.active['A1'].value = 123
```

Then, when you stop editing the cell, you'll get: 

```
  File "C:/Users/felix/Dev/xlwings/scratch.py", line 3, in <module>
    xw.sheets.active['A1'].value = 123
  File "C:\Users\felix\Dev\xlwings\xlwings\main.py", line 1570, in value
    conversion.write(data, self, self._options)
  File "C:\Users\felix\Dev\xlwings\xlwings\conversion\__init__.py", line 42, in write
    pipeline(ctx)
  File "C:\Users\felix\Dev\xlwings\xlwings\conversion\framework.py", line 63, in __call__
    stage(*args, **kwargs)
  File "C:\Users\felix\Dev\xlwings\xlwings\conversion\standard.py", line 61, in __call__
    self._write_value(ctx.range, ctx.value, scalar)
  File "C:\Users\felix\Dev\xlwings\xlwings\conversion\standard.py", line 49, in _write_value
    rng.raw_value = value
  File "C:\Users\felix\Dev\xlwings\xlwings\main.py", line 1175, in raw_value
    self.impl.raw_value = data
  File "C:\Users\felix\Dev\xlwings\xlwings\_xlwindows.py", line 744, in raw_value
    self.xl.Value = data
  File "C:\Users\felix\Miniconda3\lib\site-packages\win32com\client\dynamic.py", line 565, in __setattr__
    self._oleobj_.Invoke(entry.dispid, 0, invoke_type, 0, value)
pywintypes.com_error: (-2147352567, 'Exception occurred.', (0, None, None, None, 0, -2146777998), None)

```

njwhite:
@fzumstein lots of places [create `Dispatch` objects](https://github.com/xlwings/xlwings/blob/fa6018ade4d7ca3e27c2032232a038614bcfed6b/xlwings/udfs.py#L270) without wrapping them so I've pushed a commit to make the constructors more defensive.

fzumstein:
@njwhite how can I replicate a situation that fails with the current implementation?

njwhite:
Didn’t you just have one in your previous comment? The COM retry wrapper is barely used as it is - if you replace the implementation of _ensure_com_wrapped with just an assert-instanceof check it triggers just about everywhere.

 Try running the usual test suite on a very slow file share to see if that makes it more obvious? That makes Excel block for long periods when doing IO and so is more likely to reject xlwings COM calls, I think.

fzumstein:
right, but this case is properly handled on master - I'd be interested in replicating a case that fails on master.

fzumstein:
@njwhite  I'd suggest #1398 instead, simply re-introducing `CDispatch`. Let me know what you think.

njwhite:
@fzumstein adding CDispatch is a good idea but you still need the defensive checks in the constructor or the synchronous COM retry object will silently get dropped in some places.

fzumstein:
@fzumstein Are you able to replicate the issues with #1398?

fzumstein:
I am closing this for now as I merged your original PR fixed for CDispatch. We can get back to this any time when I can replicate the issue.

Issue #1396, opened 2020-08-26
conda version check fails with space in path


fzumstein:
see #1389

PR #1397, opened 2020-08-26
Prevent zombie processes after running app.quit()

fzumstein:
Currently, running the following leaves a zombie process:

```
>>> app = xw.App()
>>> app.quit()
```

This is only cleaned up after properly quitting the python process, leaving it around hanging forever if e.g. the command prompt is closed directly.

### DIFF
--- a/xlwings/_xlwindows.py
+++ b/xlwings/_xlwindows.py
@@ -313,6 +313,10 @@ class App:
             self._xl = get_xl_app_from_hwnd(self._hwnd)
         return self._xl
 
+    @xl.setter
+    def xl(self, value):
+        self._xl = value
+
     api = xl
 
     @property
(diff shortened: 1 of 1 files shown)


wkschwartz:
The [changelog for this item](https://docs.xlwings.org/en/stable/whatsnew.html#v0-20-5-aug-27-2020) links #1396 instead of #1397

fzumstein:
@wkschwartz, fixed, thanks!

czluo:
I am using xlWings Version: 0.20.8. I ran the code above and found that the zoombie process still existed. I have to run "app.kill()" instead to end the process. Is this problem coming up again?

fzumstein:
please open an issue with a code sample to replicate.

PR #1398, opened 2020-08-28
Fix COMRetryObjectWrapper II

fzumstein:
replacing #1395


yzgeqiping:
Why do I get an error when using "@xw.ret(expand='table')"?    ，AttributeError:Range.CLSID   raise AttributeError("%s.%s"(self._username_,attr))  File......

fzumstein:
@yzgeqiping that's been introduced with 0.20.0 - opened a new issue for this:  #1404

fzumstein:
@yzgeqiping see solution under #1404

PR #1434, opened 2020-10-12
introduced "caller" argument for UDFs

fzumstein:
So far, the easiest way to get to the caller object in the form of an xlwings range object was along these lines:

```python
import xlwings as xw
from xlwings import xlplatform

@xw.func
@xw.arg('xl', vba='Application')
def get_caller_address(xl):
    return xw.Range(impl=xlplatform.Range(xl=xl.Caller)).address
```

This can now be done by introducing the `caller` argument:

```python
import xlwings as xw

@xw.func
def get_caller_address(caller):
    return caller.address
```

On the Excel side, argument with the name `caller` will be hidden, so this function will have 0 arguments. That's a breaking change, but should have minimal impact as I don't expect many users to have an existing argument called `caller`. In any case, it's easily fixable by renaming existing arguments into something else.

### DIFF
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -425,6 +425,13 @@ def call_udf(module_name, func_name, args, this_workbook=None, caller=None):
     is_dynamic_array = ret_info['options'].get('expand')
     xw_caller = Range(impl=xlplatform.Range(xl=caller))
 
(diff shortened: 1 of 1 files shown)


Colin-b:
Nice ! I had to add the application as a parameter and retrieve caller on the application. 

Does it means it will also work for async threading mode ?

PR #1444, opened 2020-10-30
Remove comtypes requirement

### LINKED ISSUE #1443: Drop comtypes requirement
On Windows, xlwings has required comtypes since at least version v0.9.0 (588f0f5634eeefeb) until at least the current master (dceeedd). After looking at it carefully, I think this may be unnecessary. The cost of requiring an unnecessary library is added time in installation under normal set ups and added complexity and size in packaging and distribution (see, e.g., #625, which is quite similar to my use case). Very shortly I will send a PR to accomplish my recommended changes.

The only imports from comtypes are `IUnknown` and `IDispatch`:

https://github.com/xlwings/xlwings/blob/dceeeddff6411f59593f72183b592346ce946e87/xlwings/_xlwindows.py#L29-L30

# `IUnknown`

`IUnknown` is only used in `comtypes_to_pywin()` in case the `interface` argument is not set:

https://github.com/xlwings/xlwings/blob/dceeeddff6411f59593f72183b592346ce946e87/xlwings/_xlwindows.py#L196-L202

But `comtypes_to_pywin` is only called once, in `get_xl_app_from_hwnd()`, with the `interface` argument set to `IDispatch`.

https://github.com/xlwings/xlwings/blob/dceeeddff6411f59593f72183b592346ce946e87/xlwings/_xlwindows.py#L217-L225

Thus

https://github.com/xlwings/xlwings/blob/dceeeddff6411f59593f72183b592346ce946e87/xlwings/_xlwindows.py#L200-L202

could be replaced as

```python
    return _PyCom_PyObjectFromIUnknown(ptr, byref(IDispatch._iid_), True)
```

with `comtypes_to_pywin` no longer taking an `interface` argument.

So the import of **`IUnknown` can be removed completely** without changing a thing.

# `IDispatch`

`IDispatch` is used only in `get_xl_app_from_hwnd()` as shown above and from `accessible_object_from_window()`:

https://github.com/xlwings/xlwings/blob/dceeeddff6411f59593f72183b592346ce946e87/xlwings/_xlwindows.py#L188-L193

(`accessible_object_from_window()` is called only from `get_xl_app_from_hwnd()` as shown above and from `is_hwnd_xl_app()`, where nothing is done with the returned pointer.)

So it looks like there is one use of `ctypes.POINTER(IDispatch)` and there are two uses of `ctypes.byref(IDispatch._iid_)`.

## `ctypes.POINTER(IDispatch)`

Correct me if I'm wrong, but I believe this declares a pointer to a structure, and, from the [code defining `IDispatch`](https://github.com/enthought/comtypes/blob/31d05e55ed67ebcd6e48fe52deb5bfea0c4cfb2f/comtypes/automation.py#L677-L694), it looks like the structure has no [`_fields_`](https://docs.python.org/3/library/ctypes.html#ctypes.Structure._fields_). `IDispatch` inherits from our friend `IUnknown`, which also [defines](https://github.com/enthought/comtypes/blob/31d05e55ed67ebcd6e48fe52deb5bfea0c4cfb2f/comtypes/automation.py#L677-L694) no `_fields_`. (The metaclass for `IUnknown` is defined [here](https://github.com/enthought/comtypes/blob/31d05e55ed67ebcd6e48fe52deb5bfea0c4cfb2f/comtypes/__init__.py#L221-L823); notice that `IUnknown._case_insensitive_` is set to `False`.)

Moreover, the pointer produced by `POINTER(IDispatch)` is only ever passed directly to `oleacc.AccessibleObjectFromWindow` (shown above) and to [`pythoncom.PyCom_PyObjectFromIUnknown`](https://github.com/mhammond/pywin32/blob/2a148e1f982d3bc7fe4b424fd733b7ae2315af78/com/win32com/src/include/PythonCOM.h#L249-L262). Being C++ functions and being called through ctypes suggests to me that `AccessibleObjectFromWindow` and `PyCom_PyObjectFromIUnknown` really only care about the memory address contained in the pointer and, for the `byref` use, the memory address at which it is stored, not any of its Python metadata it might gain by coming from `POINTER(IDispatch)` rather than [`ctypes.c_void_p`](https://docs.python.org/3/library/ctypes.html#ctypes.c_void_p). I.e., it's not like there's any type checking happening.

So I would argue that `POINTER(IDispatch)()` can be replaced with `ctypes.c_void_p()` and a comment linking to [`IDispatch`'s documentation](https://docs.microsoft.com/en-us/windows/win32/api/oaidl/nn-oaidl-idispatch).
 
## `ctypes.byref(IDispatch._iid_)`

`IDispatch._iid_` is [defined here](https://github.com/enthought/comtypes/blob/31d05e55ed67ebcd6e48fe52deb5bfea0c4cfb2f/comtypes/automation.py#L678). Can this be replaced with the following? It's just the essentials from [here](https://github.com/enthought/comtypes/blob/31d05e55ed67ebcd6e48fe52deb5bfea0c4cfb2f/comtypes/GUID.py#L1-L35).

```python
import ctypes

class _GUID(ctypes.Structure):
    # https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/49e490b8-f972-45d6-a3a4-99f924998d97
    _fields_ = [
        ("Data1", ctypes.c_ulong),
        ("Data2", ctypes.c_ushort),
        ("Data3", ctypes.c_ushort),
        ("Data4", ctypes.c_byte * 8)]

_IDISPATCH_GUID = GUID()
ctypes.oledll.ole32.CLSIDFromString("{00020400-0000-0000-C000-000000000046}", ctypes.byref(_IDISPATCH_GUID))
```

Then the two instances of `byref(IDispatch._iid_)` could be replaced with `byref(_IDISPATCH_GUID)`.

wkschwartz:
Fixes #1443

What's missing:
- A news item for `docs/whatsnew.rst` because I'm not sure how you handle in-development items.
- I didn't touch the `docs/locales/**/*.po` files that popped up with `git grep --files-with-matches comtypes` because they look old and I don't know how to use them anyway.
- I haven't added or changed any tests because what's to test?
- I haven't run the tests because I'm on my mac today and I'm hoping your CI will just do it.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -49,7 +49,7 @@ To call Excel from Python, you don't need an add-in. Also, you can use a single 
 Dependencies
 ------------
 
-* **Windows**: ``pywin32``, ``comtypes``
+* **Windows**: ``pywin32``
 
 * **Mac**: ``psutil``, ``appscript``
 
--- a/docs/license.rst
+++ b/docs/license.rst
@@ -249,34 +249,6 @@ LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
 NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. 
 
-comtypes (Windows only)
-***********************
-
-This software is OSI Certified Open Source Software.
-OSI Certified is a certification mark of the Open Source Initiative.
-
-Copyright (c) 2006-2013, Thomas Heller.
-Copyright (c) 2014, Comtypes Developers.
-All rights reserved.
-
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
-
-The above copyright notice and this permission notice shall be included in
-all copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-THE SOFTWARE.
-
 psutil (macOS only)
 *******************
 
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ if sys.platform.startswith('win'):
         pywin32 = 'pywin32 >= 224'
     else:
         pywin32 = 'pywin32'
-    install_requires = ['comtypes', pywin32]
+    install_requires = [pywin32]
     # This places dlls next to python.exe for standard setup and in the parent folder for virtualenv
     data_files = [('', glob.glob('xlwings*.dll'))]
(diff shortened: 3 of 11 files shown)


fzumstein:
This looks awesome! Yes, definitely a good move also in light of [this](https://github.com/enthought/comtypes/issues/216). There are no unit tests for this functionality. Instead, the following has to be tested manually (which I did and I couldn't find any issues):

* Call `RunPython` with an Excel workbook from an untrusted location, e.g. downloaded from the Internet
* Open the same workbook in two instances of Excel and make sure `RunPython` still works on the correct workbook

wkschwartz:
Great. What remains to be done before merger? Anything for `docs/whatsnew.rst` or `docs/locales/**/*.po`?

fzumstein:
Nothing to be done, I'll take care of that before release.

Issue #242, opened 2015-10-03
python->Excel error vs user message 


sdementen:
When an exception occurs in python code, the error message appears as a popup box in Excel (when calling python from Excel).
In my xlwings workflow, I need in fact two type of message popping to the user:
- unhandled exceptions ==> this should popup in excel like the current message (type vbCritical). Moreover, in this case, I generate also a report with a full stack trace (and the name/values of all variables in each stack) that contains enough debugging information for me to debug (on top of it, it sends it to me by email so I get this debug message from all users of the excel tool).
- handled exceptions but that should display a friendly "error message" to the user ==> I currently raise a specific exception "ExcelUserMessage". In excel, if i see such error, the popup display the message but a bit differently (type vbInformation).

It could be useful to structure a bit this workflow in xlwings as I think it could benefit users in general.
In summary:
- create specific xlwings exception : UserDataValidationError, UserGenericMessage, ... that can be raised in python and that would be displayed specifically in Excel throuhg a msgbox
- for unhandled exceptions, allow the generation of a full stack trace that would be available either in the clipboard (if ctrl-C) or in a file (that could be opened easily form excel) or managed otherwise (via email or other adhoc procedure). This would be configurable in xlwings, one could provide a "handle_error" callback that could be used to customise the error reporting mgt (like send an email).

Just some ideas but ready to provide some code (like the full stack trace with variables names/values).

fzumstein:
Sure code is always helpful!

fzumstein:
see https://github.com/xlwings/xlwings/pull/1605

PR #1605, opened 2021-05-29
User errors in Sub displayed as MsgBox (#242)

sdementen:
proposition for implementation of #242 to define a UserException (inheriting from a COMException) that can be caught in the Sub to display a nice error message (other exceptions are still breaking the VBA code with a debug dialog)

### DIFF
--- a/xlwings/__init__.py
+++ b/xlwings/__init__.py
@@ -23,6 +23,14 @@ class LicenseError(Exception):
     pass
 
 
+class UserException(win32com.server.exception.Exception):
+    def __init__(self, message, title=None):
+        super().__init__(
+            description=message,
+            scode=winerror.E_FAIL,
+            source=self.__class__.__name__ if title is None else title,
+        )
+
 # API
 from .main import App, Book, Range, Chart, Sheet, Picture, Shape, Name, view, load, RangeRows, RangeColumns
 from .main import apps, books, sheets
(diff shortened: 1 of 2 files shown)


sdementen:
I am wondering if the code for the Function could be changed from
```
failed:
    name_of_function = Err.Description
```
to something like
```
failed:
        If Err.Number = -2147418113 Then
            name_of_function = Err.Source & ":" & Err.Description
        Else
            name_of_function = Err.Description
        End If 
```
so that for UserException we would get a consistent message (vs Sub) and a bit more descriptive (with the Err.Source containing the title of the UserException.

fzumstein:
Welcome back :) I'll take a look when I get a chance

fzumstein:
Now that we have [app.alert()](https://docs.xlwings.org/en/stable/api.html#xlwings.App.alert), I guess you can just do:

```python
try:
    ...
except Exception as e:
    myapp.alert("User-friendly message")
```

which works across all platforms.

sdementen:
The app.alert is a nice addition!
Yet for handling exceptions we want to be able to raise in different function and have their message nicely formatted for the user, it is not enough. Maybe a decorator (or an extra argument to the sub/func decorators) could handle this "exception to app.alert" behavior...

fzumstein:
I am closing this as its implementation is too specific (only Sub decorators on Windows). The following works across all platforms and implementations: 

```python
import xlwings as xw

@xw.sub
def main():
    try:
        hello()
    except UserException:
        xw.Book.caller().app.alert("Oops, a UserException happend!")
    except Exception as e:
        xw.Book.caller().app.alert(str(e))


def hello():
    raise UserException("Something happened!")

```

See also https://github.com/xlwings/xlwings/issues/2125

PR #1658, opened 2021-07-18
Added app.suspend() context manager

fzumstein:
Closes issue #254 and replaces PR #260.

6 years later, the time has finally come for your PR, @sdementen 😆  Some explanations:

* I changed `freeze` to `suspend` because when a program `freezes`, it's generally a bad thing (as someone mentioned on the issue). `suspend` is used by Microsoft in OfficeScript: https://docs.microsoft.com/en-us/office/dev/add-ins/excel/performance#suspend-screen-updating

* I replaced the kwargs/boolean approach with positional string args as setting a boolean value opposite to it's effect is just too confusing for me (technically, it would be correct, no doubt):

  ```python
  with app.suspend(interactive=True):
      # This sets app.interactive = False
  ```

  Instead, this now reads:

  ```python
  with app.suspend('interactive'):
      # This sets app.interactive = False
  ```

~* The default suspends `screen`, `alerts`, `calculation` and `events`. Not sure if that's the best selection, but I definitely want to leave out `interactive` unless explicitly requested.~

Let me know what you think @sdementen and maybe @wkschwartz, you want to chime in as you liked the original PR back then.

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -77,6 +77,21 @@ class TestApp(TestBase):
         self.app1.display_alerts = True
         self.assertTrue(self.app1.display_alerts)
 
+    def test_enable_events(self):
+        self.app1.enable_events = False
+        self.assertEqual(self.app1.enable_events, False)
+
+        self.app1.enable_events = True
+        self.assertTrue(self.app1.enable_events)
+
+    @unittest.skipIf(sys.platform.startswith('darwin'), 'app.interactive is not supported on macOS')
+    def test_interactive(self):
+        self.app1.interactive = False
+        self.assertEqual(self.app1.interactive, False)
+
+        self.app1.interactive = True
+        self.assertTrue(self.app1.interactive)
+
     def test_calculation_calculate(self):
         sht = self.wb1.sheets[0]
         sht.range('A1').value = 2
--- a/tests/test_app_suspend.py
+++ b/tests/test_app_suspend.py
@@ -1,0 +1,82 @@
+import sys
+from pathlib import Path
+
+import pytest
+import xlwings as xw
+
+this_dir = Path(__file__).resolve().parent
+
+
+@pytest.fixture(scope="module")
+def app():
+    with xw.App(visible=False) as app:
+        yield app
+
+
+def test_suspend_empty(app):
+    assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating),
+            ('automatic', True, True, True))
+    with app.suspend():
+        assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating),
+                ('automatic', True, True, True))
+    assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating),
+            ('automatic', True, True, True))
+
+
+def test_suspend_screen(app):
+    assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating),
+            ('automatic', True, True, True))
+    with app.suspend('screen'):
+        assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating) ==
+                ('automatic', True, True, False))
+    assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating),
+            ('automatic', True, True, True))
+
+
+def test_suspend_events(app):
+    assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating),
+            ('automatic', True, True, True))
+    with app.suspend('events'):
+        assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating) ==
+                ('automatic', False, True, True))
+    assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating),
+            ('automatic', True, True, True))
+
+
+def test_suspend_alerts(app):
+    assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating),
+            ('automatic', True, True, True))
+    with app.suspend('alerts'):
+        assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating) ==
+                ('automatic', True, False, True))
+    assert ((app.calculation, app.enable_events, app.display_alerts, app.screen_updating),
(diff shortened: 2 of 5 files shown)


sdementen:
Great Felix!
I like the API. 
Regarding the default, what is the behavior if you add one feature? It replaces all the default features? Or it is added to them? I feel the logic behind the default may not be obvious but not 100% sure how it works.

Other question, how do you unsuspend some feature? If you embed two suspend context manager where the first suspend everything and you want to unsuspend something in the inner context, how you can unsuspend only one of the features?

While I completely agree with you on double negation in booleans being troublesome I find the ability to assign True/False/None quiet convenient to have fine control over the features to suspend (True/False=activate/deactivate, None='do not change).

wkschwartz:
I can take a closer look later but initial reaction is what about passing the function sets of `enum.Enum` values to turn on or off instead of a string? (Or better yet, `enum.Flag`s that users can `|` together. The keyword True/False/None approach would be fine too.) I tend to prefer APIs that can be described in terms of stricter types, but I know that's a controversial point of view in Python. My feeling is type-based APIs are more self documenting and less error prone than just throwing strings at a problem. Tools often have more useful hints to offer when using types than strings. Anyway, it's up to you. Just a thought.

fzumstein:
Agreed on the defaults @sdementen, that was a bad idea and it's gone now (I was confused again by the defaults on your PR as that one worked with True/False, not with None). So currently, everything is like your PR, except the name of the function and args instead of kwargs.

@wkschwartz I had thought of Enums previously, so that might be a good compromise. How would you envision the API in that case? something like the following?

```python
with app.suspend(Suspend.INTERACTIVE, Suspend.SCREEN):
    # This sets app.interactive = False
```

wkschwartz:
Or something with [`enum.Flag`](https://docs.python.org/3/library/enum.html#flag):

```python
import xlwings as xw
# ...
with  app.interface(
    play=xw.Interface.INTERACTIVITY | xw.Interface.SCREEN,
    pause=xw.Interface.CALCS | xw.Interface.ALERTS
):
    # ...do some work...
```

(I'm not married to particular spellings. `CALCS` seems clear to me and is shorter than `CALCULATIONS`, but whatever.) The semantics I'd expect from that is that inside the `with` block, interactivity and screen are guaranteed to be "on" and calculations and alerts are guaranteed to be off. After the with block, all four things would return to their previous values.

The one thing I don't love about doing all this in a context manager is that context managers imply guarantees about things that happen inside and after `with` block, but interfacing with Excel from Python through COM is inherently race-y. If another process (or the user) turns interactivity off while the code example above is inside the `with` block, what should happen after the `with` block? Return to the value of interactivity before the `with`, potentially stomping on the other process? Doesn't seem great. A non-context-manager API would make it clearer to the programmer what xlwings can and cannot guarantee. I'm thinking something like this:

```python
previous_state =  app.interface(
    play=xw.Interface.INTERACTIVITY | xw.Interface.SCREEN,
    pause=xw.Interface.CALCS | xw.Interface.ALERTS
)
# ...do some work...
# Restore previous state *except CALCs* (in case another process did something...? not sure here.)
previous_state.play &= ~xw.Interface.CALCS
app.interface(previous_state)
```

Either way, `app.interface` would take either two named arguments or an `InterfaceState`, which is just a dataclass with the same names for instance attributes. In the first version of the API, `app.interface` returns a context manager. In the second, it returns an `InterfaceState`.

Or something like that.

wkschwartz:
If you hate `enum.Flag`, you could do `set`s of `enum.Enum` values, but I'd prefer flags.

sdementen:
I like very much your play/pause (could also be named enable/disable).
For the context manager restoring the previous state, I think it is an essential feature. Yet, we could have more control on the restoring method by having an extra argument with values "none" (no restoration), "full" (restore the state as seen when entering the context), "delta" (revert only the changes done when entering the context block)

fzumstein:
Thanks for all your inputs! I'll go through it with a few more people before releasing this, but it may take a few weeks (holiday season). In the meantime, I'd be interested if you can provide some actual use cases from your past code. For example, my use case is this: https://github.com/xlwings/xlwings/blob/master/xlwings/main.py#L763-L766

wkschwartz:
Also just noticed that the "interactivity" setting can't be manipulated on mac. Just having a warning, as the current code does, is kinda rough. I generally find warnings noisy at all the wrong times (except for the specific use case of interactive terminal sessions, which is very much _not_ how I use xlwings). Much better to fail with an exception.

In my proposed `xlwings.Interface` enumeration type, I would do something like this:

```python
class Interface(enum.Flag):
    CALC = enum.auto()
    # ...
    # Conditionally included enumation types should go last
    if sys.platform != "darwin":
        INTERACTIVITY = enum.auto()
        """A nice docstring....

        Not available on macOS.
        """
```

I know you would end up implementing the class in the separate mac/windows modules, but my point is simply not to include it at all on macOS, just as, say, Python doesn't include  [`msvcrt`](https://docs.python.org/3/library/msvcrt.html) on macOS.

If you hate this approach, please at least switch from a warning to raising an exception—possibly with a custom exception class (not `NotImplementedError`, which is for APIs that will exist one day).

-------
> In the meantime, I'd be interested if you can provide some actual use cases from your past code.

Will do when I get a chance.

wkschwartz:
> For the context manager restoring the previous state, I think it is an essential feature.

Then the docs just need to be really clear about the semantics of the context manager in the face of race conditions. Ideally these "race semantics" (is there a fancy comp sci term for this?) be well tested.

sdementen:
For my use cases, the race conditions are not an issue as I use the context
manager (home made) in subs (not functions) in Excel tools that are used
one at a time.
I use the context manager around operations like a) changing table filters
(to avoid flickering), b) updating multiple data tables without updating
formulas (for speed), c) creating new books to export data (to avoid
flickering). I also disable sometime the event handling when I have
worksheets with VBA code that track changes/activations.

fzumstein:
Awesome, thanks again for all infos, I shall be back with a new PR!

fzumstein:
replaced by https://github.com/xlwings/xlwings/pull/1671

PR #1671, opened 2021-07-22
app.properties() context manager

fzumstein:
Closes issue #254 and replaces PR #260 and PR #1658.

And then you wake up one morning and have the solution:

Screen Shot 2021-07-22 at 12 12 32

Feel free to comment, @sdementen and @wkschwartz.


sdementen:
As my 4y daughter would say, "I don't like it, I love it!"
Maybe an alternative name for properties could be "app.set(display_alerts=False)" , it would be easier to remember (for me)

fzumstein:
Great! It was worth the wait then :D

`app.set()` or `app.set_properties()` would be other options, yes. I'll think about it.

sdementen:
It would be worth to replace the kwargs signature by the explicit properties that can be set  to improve documentation and ide autocompletion

fzumstein:
Well, I am merging this as-is, as:

* kwargs: true, but it also means I need to maintain the kwargs manually every time there is a new/changed app property. You can just start with `app.` to get autocomplete, then delete `app.` again. This could still be added later without it being a breaking change.
* name: I did get some external opinions on this and while `properties` may be far from ideal, `set` could give you the wrong impression that you can actually use this outside of a context manager to set properties (while nothing will happen instead).

yzgeqiping:
Hello! When I use the latest version 0.24.3, why other functions can be used successfully, but the UDF has the following problems: "no command specified in the configuration, not autostart server 1000".

PR #1714, opened 2021-09-08
PDF export for charts and ranges and PDF export quality

### LINKED ISSUE #1697: to_pdf: allow to choose the quality
this is currently hardcoded to PrintQuality.

### LINKED ISSUE #1708: add to_pdf() support for Chart and Range

FabianGoessling:
Hey, this pull request adds:
- Pdf export for Charts and Ranges
- Quality can be set for all to_pdf methods by a new parameter "quality"
- The new parameter should not brake any existing code as the parameter is optional and has default value xlQualityStandard

I currently don't have access to a Windows machine /w Excel, so there might still be some bugs (will check for that the next days).

On the Mac Version (where i'm developing) both features don't have any effect as the 

```
object.api.ExportAsFixedFormat
```

method is not available via Appscript (might be a MacOS "Feature" as well..)

Happy for feedback, so tell me what you think..

Fixes #1708 and fixes #1697

### DIFF
--- a/xlwings/_xlmac.py
+++ b/xlwings/_xlmac.py
@@ -348,7 +348,8 @@ class Book:
     def activate(self):
         self.xl.activate_object()
 
-    def to_pdf(self, path):
+    def to_pdf(self, path, quality=None):
+        # quality parameter for compatibility
         hfs_path = posix_to_hfs_path(path)
         display_alerts = self.app.display_alerts
         self.app.display_alerts = False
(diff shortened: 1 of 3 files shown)


fzumstein:
thanks! will get to it hopefully by the end of the week!

fzumstein:
Replaced by #1808. Sorry that I had this lying around for so long. Changes I made: 

* strings instead of Classes for quality param (so it fits with the rest of the library)
* make all options available in to_pdf (show, layout)
* deduplicate code by moving to_pdf to utils
* added args to doc strings
* Range doesn't require to be selected before exporting to PDF

For next time it would be great if you could open 1 PR for 1 issue instead of combining. Thanks!

PR #1787, opened 2021-12-15
macro code documentation formatting

Jeroendevr:
Related to #1754, it had one return to many

### DIFF
--- a/xlwings/main.py
+++ b/xlwings/main.py
@@ -480,7 +480,6 @@ class App:
         However typed arrays are not supported. So the following won't work
         
         .. code-block:: vb.net
-
         Function MySum(arr() as integer)
             ''' code here
         End Function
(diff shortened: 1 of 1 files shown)


xlwings/main.py
```diff
@@ -482,7 +483,7 @@ def macro(self, name):
         .. code-block:: vb.net
 
         Function MySum(arr() as integer)
-            ''' code here
+            # code here
```
fzumstein, line 486: given this is VBA, `'` would make more sense?

Jeroendevr: Yes I thought so, but the formatting at https://docs.xlwings.org/en/stable/api.html#xlwings.App.macro was off and I thought perhaps the ‘#’ was giving the formatting error.

fzumstein: I see, no, it's the missing indentation.

fzumstein:
replaced by 942bab2bdcd93774c603f1e88e3caacfc3c4bbec

PR #1956, opened 2022-07-08
change App.range, make it consistent with Sheet.range

752377941:
change App.range, make it consistent with Sheet.range


fzumstein:
@752377941 I've replaced this PR with #1982 -- can you please confirm that it works for you?

PR #2100, opened 2022-11-21
Enable VBA functions to call async functions (async_mode='threading)

jerronl:
Previously such function can only called by the udf from the spreadsheet. After this change we can also call such functions in the VBA without waiting for the call to finish.

### DIFF
--- a/xlwings/udfs.py
+++ b/xlwings/udfs.py
@@ -82,6 +82,17 @@ async def async_thread(base, my_has_dynamic_array, func, args, cache_key, expand
     except:  # noqa: E722
         exception(logger, "async_thread failed")
 
+async def async_thread_nocaller(func, args,):
+    backcompat_check_com_initialized()
+    try:
+        loop = get_running_loop()
+        await loop.run_in_executor(
+            com_executor, functools.partial(func, *args)
+        )
+
+    except:
+        exception(logger, "async_thread failed")
+
 
 def func_sig(f):
     s = inspect.signature(f)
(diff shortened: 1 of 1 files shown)


fzumstein:
Looks interesting! Thanks!

jerronl:
Do you like this change? I accidentally updated xlwings with conda yesterday and all my workbooks stopped working until I figured and manually applied this patch.

fzumstein:
Thanks for bumping! I would like to remove the Python 3.6 hacks from back then first (now that we support only Python 3.8+) before introduce more code that depends on it. Basically reverting this commit: https://github.com/xlwings/xlwings/pull/1318/commits/ad0c9220afcac64cff3dfc51d438f7a680c19852

jerronl:
Thanks for your reply!

fzumstein:
I've now upgraded the asyncio code -- do you mind to rebase from main and update your PR?

jerronl:
I merged my branch. Is it good now? not sure what I need to do

xlwings/udfs.py
```diff
@@ -59,6 +59,17 @@ async def async_thread(base, my_has_dynamic_array, func, args, cache_key, expand
     except:  # noqa: E722
         exception(logger, "async_thread failed")
 
+async def async_thread_nocaller(func, args,):
+    backcompat_check_com_initialized()
```
fzumstein, line 63: this doesn't exist now anymore

jerronl: Sorry, I was not familiar with the pull interface in VSCode yet. How about now?

fzumstein: Looks good now, thanks! will give this a look

Issue #86, opened 2014-10-10
Add proper support for datetime.time objects/ time-only cell formats


fzumstein:
On Mac, this is taken care of by appscript, but on Windows, `_datetime_to_com_time` doesn't handle this yet.

fzumstein:
Excel to Python returns a float in any case, note a datetime object.

rudolfbyker:
I'm looking to return `datetime.time` objects from Python to VBA. I just want to check, is this the right issue to address that?
```py
@xw.func
def test_time():
    return datetime.time(1, 2, 3)

@xw.func
def test_date():
    return datetime.date(2001, 2, 3)

@xw.func
def test_datetime():
    return datetime.datetime(2001, 2, 3, 4, 5, 6)
```

![image](https://user-images.githubusercontent.com/10025342/216971430-28a769e4-5178-4061-be65-501e207aae9d.png)

fzumstein:
Basically, the issue is that Excel doesn't have a time-only type. A time-only cell has the date 1/1/1900, so you should be able to send over a `datetime` object with 1/1/1900 as the date and then assign the right `number_format` to format it as time-only.

rudolfbyker:
Thanks for the workaround! Any chance of getting that into `_datetime_to_com_time` as default behaviour? I agree it's weird, but if that's the only way Excel can represent time-only things, then that's probably what we should use.

I'm trying to make a PR but failing miserably so far. I can't figure out where the error originates.

fzumstein:
you'll also need to add `dt.time` to 
https://github.com/xlwings/xlwings/blob/main/xlwings/_xlwindows.py#L91

or it's not triggering that conversion in:
https://github.com/xlwings/xlwings/blob/main/xlwings/_xlwindows.py#L459

heetbeet:
This feature will be really helpful, thanks

PR #2162, opened 2023-02-07
#86 Support returning `datetime.time` objects from UDFs.

rudolfbyker:
In progress … The one pushed commit fixes my use case, but I'm still trying to figure out how testing works in this repo.

### DIFF
--- a/xlwings/_xlwindows.py
+++ b/xlwings/_xlwindows.py
@@ -88,7 +88,7 @@ except ImportError:
     PIL = None
 
 
-time_types = (dt.date, dt.datetime, pywintypes.TimeType)
+time_types = (dt.date, dt.time, dt.datetime, pywintypes.TimeType)
 if np:
     time_types = time_types + (np.datetime64,)
 
@@ -395,6 +395,18 @@ def _datetime_to_com_time(dt_time):
             dt_time.year,
             dt_time.month,
             dt_time.day,
+            tzinfo=win32timezone.TimeZoneInfo.utc(),
+        )
+
+    if type(dt_time) is dt.time:
+        # Excel does not have a time-only data type. It stores times with date `1900/01/01`.
+        dt_time = dt.datetime(
+            1900,
+            1,
+            1,
+            dt_time.hour,
+            dt_time.minute,
+            dt_time.second,
             tzinfo=win32timezone.TimeZoneInfo.utc(),
         )
 


rudolfbyker:
@fzumstein I'm currently following `DEVELOPER_GUIDE.md`, and getting stuck at `pip install -e ".[all]"` due to some issue with Rust. I'm not a Rust developer (yet?) so is there any way to just get tests running in the xlwings repo without having to install Rust?

If this is too complicated, would you consider writing the tests for this PR? I'm pretty sure my code will work as is.

fzumstein:
Try `python setup.py develop` instead.

rudolfbyker:
Thanks. I followed that, and the rest of the instructions. I wrote my test UDFs in `udf_tests.py`, and added them in `udf_tests.xlsm`. But I'm still not managing to run them. When I click "Import Functions", it hangs for a few minutes, and then stops with this error:

```
Could not activate Python COM server, hr = -2147221164 1000

Press Ctrl+C to copy this message to the clipboard.
```

I searched for this message, and found two things that don't look applicable to my situation:
- Using Microsoft Office from Store (No, I'm using Office 365)
- Empty lines in config file (Not using a config file as far as I can tell)

Any ideas?

fzumstein:
If you check `Show Console` you should probably get a better error message

rudolfbyker:
In the console, I see:

```
  File "<string>", line 1, in <module>
  File "c:\aa\xlwings\xlwings\server.py", line 16, in <module>
    raise ImportError(
ImportError: Couldn't find 'pywin32'. Install it via'pip install pywin32' or 'conda install pywin32'.
```

However, `pywin32` is definitely installed in the `venv` that I'm using, and I have the interpreter set to `C:\aa\xlwings\venv\Scripts\python.exe` in the ribbon (I have the git repo cloned at `C:\aa\xlwings`).

After typing `exit` in the console, I get a popup saying:

```
Python process exited before it was possible to create the interface object.
Command: cmd.exe /showconsole /K ""C:\aa\xlwings\venv\Scripts\python.exe" -B -c "import sys, os;import xlwings.utils;xlwings.utils.prepare_sys_path(\"true;C:\\aa\\xlwings\\tests\\udfs\\udf_tests.xlsm;C:\\aa\\xlwings\\xlwings\\addin\\xlwings.xlam;;;;\");import xlwings.server; xlwings.server.serve('{d5eab099-116e-462c-9ece-8c975870209f}')""
Working Dir: <unspecified> 1000
```

I have no idea what to make of this.

fzumstein:
In this case, you've got an issue with pywin32 as `import win32api` fails: https://github.com/xlwings/xlwings/blob/main/xlwings/server.py#L14

rudolfbyker:
> In this case, you've got an issue with pywin32 as `import win32api` fails: https://github.com/xlwings/xlwings/blob/main/xlwings/server.py#L14

Now you're just repeating exactly what I said in the first half of my post...

PR #2219, opened 2023-04-05
pytest as dependency for development

Jeroendevr:
As the developguide of xlwings states, xlwings is in transition to pytest. Therefor installing it with the 
```shell
pip install -e ".[all]"
```
command should also install pytest.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ all = [
     "flask",
     "requests",
     "pdfrw",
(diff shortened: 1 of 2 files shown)


fzumstein:
for now, this will have to be duplicated in `setup.py`

Jeroendevr:
While we are at it. Also include the `black` formatter.  As I had formatting issues due to not using black.

PR #2223, opened 2023-04-06
docs pro - clarify workbook location

Jeroendevr:
Given the current layout for installers, I thought it's better to be explicit about the workbook location given the project layout of installers.  (if this is the correct location or should it be placed below src?)


fzumstein:
This is actually outside of the installer's repo. The idea is that you don't have to commit any of your code/workbook to the 1-click installer's repo.

Jeroendevr:
How to interpret the sentence?

> To work with the release command, you should have your workbook in the xlsm format next to your Python code:
So you assume having two locations of source code? One for the installer and one as a development repository?

I assumed based on that sentence that using the release command is related to the location of the xlsm format (same name of folder, package etc.?)

fzumstein:
Yes, the 1-click repo is only for building the installer. You can of course add other stuff, but that's up to you and users don't generally do that as I have access to the repo/it's not under their user account.

PR #2225, opened 2023-04-07
Avoid the packaged application flashing black cmdline window.

BruceLee569:
```python
import xlwings as xw
```

Package an exe application through `pyinstaller -w main.py` ,  main.py with the above code,  because the subprocess `tasklist` command is called, a black cmdline window will flash every time when the program exits.

### DIFF
--- a/xlwings/_xlwindows.py
+++ b/xlwings/_xlwindows.py
@@ -514,6 +514,7 @@ class Apps(base_classes.Apps):
             split('tasklist /FI "IMAGENAME eq EXCEL.exe"'),
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
+            shell=True,
             encoding="utf-8",
         )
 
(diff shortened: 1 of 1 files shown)


fzumstein:
I'd like to avoid introducing `shell=True`, see https://stackoverflow.com/a/3172488/918626
Is there no way to suppress this with a pyinstaller setting? Otherwise, I might think about removing the atexit handler again so it won't get called automatically upon exit of the program. Not sure if leaving it in the App.quit() and App.kill() would still cause this issue?

BruceLee569:
After a brief search, we can use `creationflags` to avoid cmdline window flash, see https://stackoverflow.com/a/2935727/11687405.

Now we just need to change `shell=True` to `creationflags=0x08000000`, or `creationflags = subprocess.CREATE_NO_WINDOW` from Python 3.7.

fzumstein:
great! As xlwings doesn't support < 3.7, I'd prefer to use `subprocess.CREATE_NO_WINDOW` though

Issue #2265, opened 2023-05-24
test_markdown.py fails due XLWINGS_LICENSE_KEY_SECRET being none


Jeroendevr:
#### OS (e.g. Windows 10 or macOS Sierra)
macOS Ventura
#### Versions of xlwings, Excel and Python (e.g. 0.11.8, Office 365, Python 3.7)
xlwings - Latest,
Python 3.11
Office 365
#### Describe your issue (incl. Traceback!)

```python
pytest tests/test_markdown.py   
```

```python
tests/test_markdown.py:8: in <module>
    from xlwings.pro import Markdown, MarkdownStyle
xlwings/pro/__init__.py:14: in <module>
    from .embedded_code import dump_embedded_code, runpython_embedded_code
xlwings/pro/embedded_code.py:25: in <module>
    LicenseHandler.validate_license("pro")
xlwings/pro/utils.py:99: in validate_license
    os.getenv("XLWINGS_LICENSE_KEY_SECRET").encode(),
E   AttributeError: 'NoneType' object has no attribute 'encode'

ERROR tests/test_markdown.py - AttributeError: 'NoneType' object has no attribute 'encode'
```

Perhaps related to #2220 and #2241?

This secret is something only xlwings owns correct? Or is it a derivative from the License key?

As for the solution, I cannot wrap this part in a Try import just like #2220 can I? Otherwise I cannot run the tests.

fzumstein:
You'd either have to exclude running tests that use the pro subpackage or use the noncommerical license key.

Jeroendevr:
Yes that helps, I've opened a PR #2266 to prevent/explain similar questions for future users.

fzumstein:
closed by 34a5302cdfcf64e7ba43f843275640096d67da70

PR #2266, opened 2023-05-24
Developer guide - xlwings_license_key_secret

Jeroendevr:
Little improvement to the developer guide, so that contributing developers now that the key has to be set to `noncommercial` instead of their actual license key for running xlwings pro related tests. 

Explains #2265

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,4 @@ node_modules
 .parcel-cache
 
 .hypothesis/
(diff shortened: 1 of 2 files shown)


fzumstein:
Replace with https://github.com/xlwings/xlwings/commit/34a5302cdfcf64e7ba43f843275640096d67da70 as the formulation isn't technically correct.

PR #2269, opened 2023-05-25
activate app before testing if it's active

Jeroendevr:
Similar to #2262. The test implicitly assumed that the last created app should also be the activated one. 

Also because of the transition to pytest, the `import unittest` is removed. Is the 
```Python
if __name__ == "__main__":
    unittest.main()
```
needed as I haven't seen it in `test_font.py` as this was mentioned as the example module transitioned to pytest.

### DIFF
--- a/tests/test_active.py
+++ b/tests/test_active.py
@@ -1,7 +1,4 @@
 import os
-import pytest
-
-
 import xlwings as xw
 
 from .common import TestBase, this_dir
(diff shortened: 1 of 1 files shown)


fzumstein:
that's cool, but can you run `isort .` and recommit?

Jeroendevr:
yes thats's it.

PR #2284, opened 2023-06-19
UDF Reserved argument "caller" tests

philipdp123:
Added in two tests to ensure the UDF reserved 'caller' argument functions as expected.

`cell_caller` tests that the caller argument is as expected

`cell_caller_multiple_args` tests that the caller argument is as expected even when multiple arguments are passed.

Tests have been added to the udf_tests.xlsm file as well.


fzumstein:
thanks!

PR #2305, opened 2023-07-20
[Feature request] Retrieve column letter #2091

### LINKED ISSUE #2091: [Feature request] Retrieve column letter
Hi,

I would like to suggest adding a function or an argument to retrieve the column letter from a given range. Retrieving the column number is possible with [.column](https://docs.xlwings.org/en/stable/api.html#xlwings.Range.column) in xlwings, but retrieving the column letter is currently not supported with built-in functions (as far as I know). It is possible with the fiddly line
`sheet["A1"].address.split("$")[1]`

However, it would be smoother to add a built-in function, for example, openpyxl has the function [.get_column_letter](https://openpyxl.readthedocs.io/en/stable/api/openpyxl.utils.cell.html#openpyxl.utils.cell.get_column_letter) (and [.column_index_from_string](.column_index_from_string)), which is used to convert a column index into a column letter, e.g. 3 -> “C” (and “C” -> 3, respectively).

I think adding an argument to [.column](https://docs.xlwings.org/en/stable/api.html#xlwings.Range.column) could be an user-friendly solution.

Cheers
mullimanko

philipdp123:
This PR aims to resolve #2091.

I noted that you had a function that achieves this task. My PR contains a reworked version of ``xw.utils.col_name`` inclusive of a few test using pytests to ensure that the function works as expected.


mullimanko:
This is great, thank you. When will this be implemented?

fzumstein:
I am not quite sure about the motivation of this PR as the function is already implemented (I didn't remember this, so forgot to mention it in https://github.com/xlwings/xlwings/issues/2091). Also keep in mind that xlwings.utils is also been more of an internal API.

The new implementation is almost twice as slow as the current implementation, so not likely an improvement?

Screenshot 2023-09-24 at 2 23 17 PM

goroderickgo:
Sorry @fzumstein , when you say "the current method", what is that? I've searched the docs many times for an easy way to get column letters or row numbers but have never been able to find it.

fzumstein:
```python
from xlwings.utils import col_name
```

goroderickgo:
ohhhh neat.

Issue #2026, opened 2022-09-23
Redesign xlwings-*.dll packaging and handling


fzumstein:
* Currently packaged via data_files, which can place data files outside of the package (in our case next to the python.exe), but it is deprecated and isn't handled by any of the new packaging tools
* It has to be packaged inside the package: Probably same solution as with applescript file on macOS, although tbd if the dlls should be copied again next to the interpreter, or if there's a nicer solution
* Will probably automatically solve https://github.com/xlwings/xlwings/issues/1187 and https://github.com/xlwings/xlwings/issues/1928

PR #2312, opened 2023-08-10
Feature request: Support both `Object` and `Variant` return types in UDFs.

rudolfbyker:
I tested this manually in our system, and it seems to work.

Sorry, no unit tests, because I could not figure out how to get tests running in this project.


fzumstein:
I added instructions on how to run the UDF tests: https://github.com/xlwings/xlwings/blob/main/DEVELOPER_GUIDE.md#tests

If no unit test, a code sample of what this PR allows you to do would be great.

fzumstein:
And ignore the pipeline failure, that's now fixed on main.

rudolfbyker:
OK, so I ran into a few problems here... Please bear with me.

> I added instructions on how to run the UDF tests: https://github.com/xlwings/xlwings/blob/main/DEVELOPER_GUIDE.md#tests

The only relevant thing I see here is:

> To run the UDF tests, open udf_tests.xlsm in the tests/udfs directory and follow the instructions at the top of the file. The other files work accordingly.

I created a venv for xlwings development. If I don't install anything, and just follow these instructions, I get `ModuleNotFoundError: No module named 'nose'`. Then I went looking for requirements.txt, setup.py or pyproject.toml files. None of them contain `nose`. When I run `pip install nose`, I get `Requirement already satisfied: nose`.

Then I noticed that xlwings is using the wrong Python. To try to fix that, I ran `xlwings addin install` from within the venv, but that gave me a dependency error.

To try to install dependencies, I ran `pip install .` inside the repo folder, using the venv, but that told me I need Rust installed...

It looks like I'm going to have to build a separate VM for xlwings development, and install a lot of dependencies. I would really like to help you here, but without better instructions on how to get started, that's almost impossible.

-----

> If no unit test, a code sample of what this PR allows you to do would be great.

Sure, here is what it allows me to do:

```vba
Public Function foo() As Collection
    Set foo = New Collection
    foo.Add "bar"
    foo.Add "baz"
End Function
```

```py
vba_collection_object = wb.macro("foo")()
```

Use cases:

- Creating VBA objects in VBA, keeping them in memory in Python, and passing them back as arguments to other VBA functions when calling them from Python.
- VBA functions that sometimes returns an object, and sometimes returns a scalar value.

fzumstein:
How to get by without Rust is explained here: https://github.com/xlwings/xlwings/blob/main/DEVELOPER_GUIDE.md#python-package

rudolfbyker:
I'm on a fresh Windows 10 installation, I followed the steps at https://github.com/xlwings/xlwings/blob/main/DEVELOPER_GUIDE.md#tests exactly (even with Rust), and I had to do two extra things which are not specified in the developer guide to get it to work:

- Specify my interpreter path in the "Interpreter" field on the xlwings ribbon, since I'm using a venv.
- `pip install nose`

Also, the correct place for the `dll`s seems to be directly in the `venv` folder, rather than next to `python.exe` (which is in `venv/Scripts`).

But the good news is that I can run the UDF tests now.

fzumstein:
* I tried to improve the docs: https://github.com/xlwings/xlwings/commit/d7e025058cf5e684adb066a9848cc45b4e63ad01
* The dll location sound good to where they are expected in a venv. However, there's an open issue to redesign this completly: https://github.com/xlwings/xlwings/issues/2026

rudolfbyker:
As I'm writing the unit tests, I realized that I gave a bogus example above of what I'm trying to do! So let me clarify the use case:

### 1. Returning arbitrary objects from VBA to Python already works

```
Public Function foo() As Collection
    Set foo = New Collection
    foo.Add "bar"
    foo.Add "baz"
End Function
```

```
vba_collection_object = wb.macro("foo")()
```

### 2. Returning objects like a dict from Python to VBA already works

- This is already covered by the `write_dict` UDF test.
- It does not do what I want it to do, so I never use this.

### 3. Returning arbitrary VBA objects from Python to VBA (not from VBA to Python)

This is what is currently not possible, and what my PR will solve.

```
@xw.func
def return_vba_dict() -> Any:
    return create_vba_dict({"a": 1, "b": 2})


def create_vba_dict(d: Mapping[str, Any]) -> Any:
    """
    Create a VBA dictionary.

    See https://stackoverflow.com/questions/67397267/pass-dictionary-to-excel-macro-using-win32com-and-comtypes
    """
    import win32com.client

    result = win32com.client.Dispatch("Scripting.Dictionary")

    for key, value in d.items():
        result[key] = value

    return result
```

Similarly, we might want to pass something returned in case 1 above (e.g. the collection object or some other object) back to VBA at some point.

Without my PR, we will get an error like this:

image

With my PR and the example above, it works, and VBA gets a bona fide `Dictionary` object back from Python.

rudolfbyker:
Rebased and added tests :)

fzumstein:
Awesome, thanks!

rudolfbyker:
Rebased again. Please consider.

fzumstein:
thanks, I'll try to get to it next week

rudolfbyker:
Rebased again. Please consider. :)

PR #2343, opened 2023-10-04
test_app refactor

Jeroendevr:
- Updated isort to work with python 3.12.0
- Removed Unittest references
- Runnable with Pytest
- Added `pytest.mark.skipif` for macOS platform
- Changed assertions to be accepted by Black

Welcoming benefit is that the test explorer in VScode now can find the tests without errors. 

Note:
the `test_iter` was failing on macOS, I assumed this was because of [this](https://github.com/xlwings/xlwings/blob/b389c15867cdf39f1a6e3ee16edb2da150d742d6/xlwings/main.py#L300C1-L306C8) Therefore I did a `skipif`


fzumstein:
hm you're running this against an old code base. on the current main branch, I can run everything as-is and don't have any issues with pytest and black (pytest can run unittests without any rewriting). It may be that the tests run better if no Excel instance is running at the time of starting the tests.

Jeroendevr:
I had a change locally from a long time ago preventing a proper rebase. Excuse me for the PR

PR #2358, opened 2023-11-29
UDF tutorial update

cxhx441:
I thought this section could be expanded upon a bit.


fzumstein:
Thanks! However, could you stick to PEP8 for module names, i.e., use `my_udf` or `myudf` rather than `MyUDF`?

cxhx441:
Thank you for the feedback! I have made the requested changes.

Issue #2335, opened 2023-09-20
xlwings vba UnicodeDecodeError stops import or edit from running.


nmadura:
#### OS (e.g. Windows 10 or macOS Sierra)
Windows 10
#### Versions of xlwings, Excel and Python (e.g. 0.11.8, Office 365, Python 3.7)
xlwings version: 0.30.12
Microsoft® Excel® for Microsoft 365 MSO (Version 2306 Build 16.0.16529.20226) 64-bit 
Python 3.11.5
#### Describe your issue (incl. Traceback!)
when encountering an invalid unicode sequence xlwings errors out. This functionality may be OK, however it provides little detail to proceed with.

```python
# Your traceback here
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\xxxx\AppData\Local\Programs\Python\Python311\Scripts\xlwings.exe\__main__.py", line 7, in <module>
  File "C:\Users\xxxx\AppData\Local\Programs\Python\Python311\Lib\site-packages\xlwings\cli.py", line 1358, in main
    args.func(args)
  File "C:\Users\xxxx\AppData\Local\Programs\Python\Python311\Lib\site-packages\xlwings\cli.py", line 849, in vba_imp
    vba_code = f.readlines()
               ^^^^^^^^^^^^^
UnicodeDecodeError: 'cp932' codec can't decode byte 0xef in position 930: illegal multibyte sequence
```

I was able to trace down the offending area by wrapping the f.readlines() call in a try block
this is replacing line 849 in cli.py

```python
        elif path.suffix in (".cls", ".frm"):
            with open(path, "r") as f:
                try: 
                    vba_code = f.readlines()
                except UnicodeDecodeError as ude:
                    if ude.end > 50:
                        offending = ude.object[ude.end - 50:ude.end]
                    else:
                        offending = ude.object[ude.start:ude.end]
                    print("UnicodeDecodeError reading <{}>, \nexception: {} \ncontext: {}".format(path, ude, offending))
                    raise ude
```

#### Include a minimal code sample to reproduce the issue (and attach a sample workbook if required!)

```python
# Your code here

```

fzumstein:
Can you provide the source or the offending character?
You could also try to do `with open(path, "r", encoding="utf-8") as f:` in your code extract above to see if that solves it.

nmadura:
[...]

PR #2380, opened 2024-01-15
"Fix UnicodeDecodeError in some VBA file handling

LeonKohli:
### Summary
This pull request addresses the issue where `xlwings` encounters a `UnicodeDecodeError` when handling VBA files with encodings other than UTF-8. The proposed changes add a multi-encoding read strategy to the `export_vba_modules`, `vba_edit`, and `vba_import` functions in `cli.py`, allowing `xlwings` to handle files with different encodings such as 'ISO-8859-1' and 'cp1252', alongside the default 'utf-8'.

### Changes
- Modified `export_vba_modules`, `vba_edit`, and `vba_import` functions to try reading files with 'utf-8', 'ISO-8859-1', and 'cp1252' encodings.
- Added error handling to raise `UnicodeDecodeError` if all encodings fail, providing clear feedback.

### Justification
These changes enhance the robustness of `xlwings` in handling VBA files with various encodings, which is a common scenario in diverse environments. It ensures smoother functionality and reduces the likelihood of runtime errors due to encoding issues.

### Testing
The modifications have been tested in various scenarios to ensure compatibility and functionality across different file encodings.

### DIFF
--- a/xlwings/cli.py
+++ b/xlwings/cli.py
@@ -734,10 +734,10 @@ def release(args):
 
 
 def export_vba_modules(book, overwrite=False):
-    # TODO: catch error when Trust Access to VBA Object model isn't enabled
-    # TODO: raise error if editing while file hashes differ
     type_to_ext = {100: "cls", 1: "bas", 2: "cls", 3: "frm"}
     path_to_type = {}
+    encodings = ['utf-8', 'ISO-8859-1', 'cp1252']
+
     for vb_component in book.api.VBProject.VBComponents:
         file_path = (
             Path(".").resolve()
@@ -747,16 +747,23 @@ def export_vba_modules(book, overwrite=False):
         if (
             vb_component.Type == 100 and vb_component.CodeModule.CountOfLines > 0
         ) or vb_component.Type != 100:
-            # Prevents cluttering everything with empty files if you have lots of sheets
             if overwrite or not file_path.exists():
                 vb_component.Export(str(file_path))
                 if vb_component.Type == 100:
-                    # Remove the meta info so it can be distinguished from regular
-                    # classes when running "xlwings vba import"
-                    with open(file_path, "r", encoding="utf-8") as f:
-                        exported_code = f.readlines()
+                    for encoding in encodings:
+                        try:
+                            with open(file_path, "r", encoding=encoding) as f:
+                                exported_code = f.readlines()
+                            break  # Successful read, break the loop
+                        except UnicodeDecodeError:
+                            continue  # This encoding didn't work, try the next one
+                    else:
+                        # All encodings failed
+                        raise UnicodeDecodeError("Failed to decode the file with any of the provided encodings.")
+                    
                     with open(file_path, "w", encoding="utf-8") as f:
                         f.writelines(exported_code[9:])
+
     return path_to_type
 
 
@@ -798,6 +805,8 @@ def vba_import(args):
 
     book = vba_get_book(args)
 
+    encodings = ['utf-8', 'ISO-8859-1', 'cp1252']
+
     for path in Path(".").resolve().glob("*"):
         if path.suffix == ".bas":
             try:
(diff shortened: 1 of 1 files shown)


fzumstein:
Thanks! Can you also provide instructions on how to reproduce this issue? See also https://github.com/xlwings/xlwings/issues/2335

LeonKohli:
Hey @fzumstein 

Thank you for following up on the pull request. Regarding the UnicodeDecodeError issue with xlwings, I encountered it while working on a large VBA project in my professional environment. Initially, I focused on resolving the issue directly as it was impacting our workflow, rather than setting up a detailed reproduction scenario.

However, I believe the problem arises when handling VBA files with specific Unicode characters that are not encoded in UTF-8, especially in large projects where various encoding standards might have been used over time

fzumstein:
were you using files that were exported outside of xlwings?

LeonKohli:
To clarify, the files I encountered the issue with were indeed part of a larger VBA project, and some of these files may have been exported or edited outside of xlwings before being reintegrated into the project. This mixed handling could be a contributing factor to the encoding discrepancies leading to the UnicodeDecodeError.

fzumstein:
Right, in this case I think it's fair to expect that the user has to use xlwings to do the initial export though. The way you have it now is just covering your specific use case, it wouldn't work for this case: https://github.com/xlwings/xlwings/issues/2335
I'd rather allow users to specify a non-utf-8 encoding via command line switch, something like:
`xlwings vba edit --encoding=cp932`
I think it could be ok to loop through `utf-8` and `locale.getpreferredencoding()` by default, as this is still generic.

zwackelfuss:
Hello,
I am new to xlwings and just installed the latest version (0.31.0). Opening a file the way described here https://github.com/xlwings/xlwings/issues/2335#issuecomment-1730156684 (Terminal vba edit --file) , the xlsm opens, but this error occurs:
```
  File "C:\Users\Heiko\AppData\Local\Programs\Python\Python39\lib\site-packages\xlwings\cli.py", line 757, in export_vba_modules
    exported_code = f.readlines()
  File "C:\Users\Heiko\AppData\Local\Programs\Python\Python39\lib\codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfc in position 354: invalid start byte
PS C:\Users\Heiko\Documents\testxl> 
```
Is there anything I can do to resolve this problem?

fzumstein:
@zwackelfuss if you can attach a sample workbook to replicate the issue, that would help. Also, if you can report your machine's locale:

image

zwackelfuss:
Thank you Felix. See my language settings (German). 
image
Odd: I opened the xlsm and deleted some simple cell values (name of persons), saved the file and opened it in xlwings.. without error. So it seems some of the names are the reason for the error. As the files contains names, I prefer sending you the file via mail instead of posting it here - hope this is ok for you.

PedroWitzel:
Hello, first time here 👋 
I had a similar issue with a Brazilian Portuguese workbook, and grabbing this modification resolved the issue.

I'm running `xlwings vba edit` to do so.
I can do further testing if you go further with this, or if you require a 'broken' workbook.

ZwilleSmutje:
I was able to get the export working when setting the reading part of cli.py>vba_edit to 

Line 867             with open(path, "r", encoding="ISO 8859-1") as f:

JumperJu:
Hey, thanks for the code! 
I was having trouble with some Spanish characters, and your fix did the trick. 👍

thirdcoffee:
Hey, I've also stumbled upon this problem working with legacy code and looked into the code.
As far as I can tell, the import routine called in vba_import (see https://github.com/xlwings/xlwings/blob/beb6853f587ffa82ca1508837ad0d8223f3acb84/xlwings/cli.py#L799) apparently expects ISO-8859-1 or Windows 1252 encoding while working with vba edit uses utf-8 encoding to read changes (see https://github.com/xlwings/xlwings/blob/beb6853f587ffa82ca1508837ad0d8223f3acb84/xlwings/cli.py#L871).
This inconsistent behaviour might lead to corrupted files, especially in projects featuring different encodings due to their history.

Could you maybe take a new look at harmonizing vba import to feature importing utf-8 encoded files directly? @fzumstein
That would be great!

geanluiz:
> Right, in this case I think it's fair to expect that the user has to use xlwings to do the initial export though. The way you have it now is just covering your specific use case, it wouldn't work for this case: #2335 I'd rather allow users to specify a non-utf-8 encoding via command line switch, something like: `xlwings vba edit --encoding=cp932` I think it could be ok to loop through `utf-8` and `locale.getpreferredencoding()` by default, as this is still generic.

Hey @fzumstein
I was facing the same issue with special characters in Brazillian Portuguese (like ç, á and ã). They would break as soon as I imported the files back from VS Code to Excel. I tried changing the encoding inside VS Code as I learned that vba would need ANSI encoding, but then I got the same error messages of #2335

I fixed it by editing cli.py with `with open(path, "r", encoding="Windows-1252") as f:` on both functions
 
I then searched for related issues so I would be able to open a pull request and after reading this one and the issue #2335 I was wondering if You could explain further how @LeonKohli 's PR wouldn't fix the problem because, as a begginer programmer, it looks to me like it is actually an upgraded version of #2364, which is limited only to 'utf-8'.

Tiesza:
For anyone having problems, try this solution https://github.com/xlwings/xlwings/issues/2654#issuecomment-4498343242

PR #2397, opened 2024-02-19
prefer preconfigured paths to searching registry paths

robertschulze:
reordered the Sharepoint Online search steps and aligned logic to OneDrive and OneDrive for Business


fzumstein:
Got it! Yes, that definitely makes sense! Thanks!

fzumstein:
Hi @robertschulze, this introduced a regression: #2413 . Do you mind having a look to see if you have an idea why this happened? Thanks.

PR #2421, opened 2024-03-19
fix the error "The truth value of an array with more than one element…

jerronl:
fix the error "The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()"


fzumstein:
Please add a sample/test where it currently fails.

jerronl:
```
@xw.func 
def test1():
    return [np.zeros((2,2)),np.zeros((1,2))]
@xw.func 
def test2():
    return np.zeros((1,2))]

```
the current xlwings will fail test1
but my fix will fail test2. :(
we might want to check the lenth (len() would not work) to decide if .all() is necessary but it seems too trivial...

PR #2445, opened 2024-05-29
xlwings Server: fix threading issue with and properly decode image_st…

fzumstein:
…ring so it works with pyscript

### DIFF
--- a/xlwings/pro/_xlremote.py
+++ b/xlwings/pro/_xlremote.py
@@ -969,7 +969,7 @@ class Pictures(Collection, base_classes.Pictures):
         # ax = df.plot(figsize=(3,3))
         # fig = ax.get_figure()
         with open(filename, "rb") as image_file:
-            encoded_image_string = base64.b64encode(image_file.read())
+            encoded_image_string = base64.b64encode(image_file.read()).decode("utf-8")
         # TODO: width and height are currently ignored but can be set via obj properties
         self.append_json_action(
             func="addPicture",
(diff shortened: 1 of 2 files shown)


gdementen:
FWIW, the fix to this issue broke my application, which uses both matplotlib, xlwings and Qt. It took me several hours to find the cause of the issue because the failure was in matplotlib code, not in xlwings.

gdementen:
Wouldn't it be possible to do that only in the "server" part, and not in utils, so that when you "import xlwings" from a Python script, you don't have such an unexpected side-effect?

fzumstein:
What's the error message you're getting?

fzumstein:
or...wha't the exact issue? are you embedding matplotlib in qt and it's requiring a different backend?

gdementen:
The error I was getting is not very interesting IMO:
```python
[...]
  File "[...]\larray_editor\utils.py", line 302, in show_figure
    figure.canvas.manager.window is not None):
AttributeError: 'FigureManagerBase' object has no attribute 'window'
```
and I did not understand what was happening until I saw the changed line in xlwings code and I realized my matplotlib figure manager has always been Qt-based and after the change in xlwings, it was not anymore.

My use case is indeed a Qt application which uses both matplotlib (and thus requires its QtAgg backend) and xlwings. This is not something everybody does... but can't this change cause problems to more classic scripts using both matplotlib (via pyplot module for example) and xlwings? I think the default backend for matplotlib is usually Qt and currently "import xlwings" changes that, which is unexpected (at least for me).

FWIW, I just fixed the issue in my code by using `matplotlib.use('QtAgg')`, but since this is a global thing, I fear this is will depend on the order of imports, so the fix might be brittle, especially because my application can also be called from within user scripts which could use xlwings too.

fzumstein:
Got it, I'll remove it from there: https://github.com/xlwings/xlwings/issues/2484

gdementen:
Thanks a lot

PR #2542, opened 2024-10-23
Fix prepare_sys_path case sensitivity for qemu shared folder

ldebs:
For some reason, mounted shared folders on a Windows guest's remote drive appear to be case-sensitive.
This pull request addresses the issue by recreating system paths with their original case, specifically for Windows.
Additionally, the `os.realpath` method does not work in this context, so the implementation uses `os.listdrives` (python >= 3.12) and `os.listdir` to reconstruct the correct path.

### DIFF
--- a/xlwings/utils.py
+++ b/xlwings/utils.py
@@ -467,7 +467,19 @@ def prepare_sys_path(args_string):
         paths += args[6:]
 
     if paths:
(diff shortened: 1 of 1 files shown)


ldebs:
It does not fix the issue. It is only fixed when I add the Excel file directory to the PYTHONPATH en var...

PR #2543, opened 2024-10-25
fix varargs for object handles

fzumstein:
2nd attempt for #2538


fzumstein:
In the end, this has been replaced with a more efficient version in https://github.com/xlwings/xlwings-server/pull/170

Issue #1963, opened 2022-07-21
migrate from watchgod to watchfiles


fzumstein:
see: https://watchfiles.helpmanual.io/migrating/
https://github.com/xlwings/xlwings/blob/main/xlwings/cli.py#L682-L687

PR #2583, opened 2025-02-21
2580 add optional dependency of watchgod

### LINKED ISSUE #2580: Add optional dependency of watchgod to pyproject.toml to enable use of uvx for running xlwings vba edit
Running

`uvx xlwings vba edit -v -f "FILE"`

Yields

`Please install watchgod to use this functionality: pip install watchgod`

If optional dependency for vba edit was added to pyproject.toml that requires watchgod I believe this would be possible:

`uvx --from 'xlwings [vba_edit]' xlwings vba edit -v -f "FILE"`

See also https://docs.astral.sh/uv/guides/tools/#requesting-extras

This would be a nice extra feature. If this would be okay I can create the PR

MEMAndersen:
Closes #2580

Simply adds a new optional dependency to pyproject.toml of the watchgod library enablings the usage of `xlwings vba edit` via uv:

uvx --from 'xlwings[vba_edit]' xlwings vba edit"

To test (before it gets merged) use the following:
`uvx --from "git+https://github.com/MEMAndersen/xlwings.git@2580-Add-optional-dependency-of-watchgod[vba_edit]" xlwings vba edit`

See also https://docs.astral.sh/uv/guides/tools/#requesting-extras

When #1963 is adressed the optional dependency should be updated

It is my first time contributing to anything so please guide or correct me if i am doing something wrong

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ all = [
     "pytest",
     "mistune",
 ]
+vba_edit = ["watchgod"]
 
 [project.scripts]
 xlwings = "xlwings.cli:main"
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ extras_require = {
         "pytest",
         "mistune",
     ],
+    "vba_edit": ["watchgod"],
 }
 
 if os.getenv("BUILD_RUST", "0") == "1":


fzumstein:
Thanks and all good! The only thing is that you currently need to also do it here: https://github.com/xlwings/xlwings/blob/main/setup.py#L44-L59 as the Windows packaging still relies on setup.py because of the dll handling.

MEMAndersen:
Thank you for a quick response Felix! I have added the optional dependency to `setup.py` as well now

PR #2609, opened 2025-04-21
add directories for vba structure

balex723:
split vba files to directories Sheets(incl.workbook)/Modules/Classes/Forms/


balex723:
I can't test this request (sorry) and thus not sure watchgod walks though subdirs of source dir.

PR #2655, opened 2025-12-03
treat CondaEnv as path if path separator found

tl-hbk:
Currently if a conda env wasn't in the correct location it would fail to load the DLL.  This is documented in the vba code

```
        ' This only works if the envs are in their default location
        ' Otherwise you'll have to add the full path for the interpreter in addition to the conda infos
```

The workaround it suggests though causes a different part of the code to fail.  If you provide the full path to the interpreter

```
    If (PYTHON_WIN = "python" Or PYTHON_WIN = "pythonw") And (CondaPath <> "" And CondaEnv <> "") Then
        CondaCmd = Chr(34) & Chr(34) & CondaPath & "\condabin\conda" & Chr(34) & " activate " & CondaEnv & " && "
        PYTHON_WIN = "cmd.exe " & ConsoleSwitch & " /K " & CondaCmd & "python"
    Else
        PYTHON_WIN = "cmd.exe " & ConsoleSwitch & " /K " & Chr(34) & Chr(34) & PYTHON_WIN & Chr(34)
    End If
```

The conda activate doesn't occur and some packages in your environment that rely on the environment being activated might not work properly.  My use case is some c extensions that need the PATH modified by conda so that some other dlls are found.


fzumstein:
Thanks! Can you remind me which case this covers, i.e., what combination of settings and conda setup?

tl-hbk:
I have `envs_dirs` set in `.condarc` to a shared location on the machine outside of the default `<conda install location>/envs`.  Instead of passing an environment name where xlwings would asssume  `<conda install location>/envs/<name>` is the path of the environment, I'd like to be able to pass the entire path so something like `D:\conda_envs\foo_env`.

This lets the xlwings vba code find the correct DLL location while still activating the python environment for the python calls.

An example `xlwings.conf` in the workbook would be

```
Interpreter_Win	python
Conda Path	C:\Users\<user>\miniconda3\
Conda Env	D:\conda_envs\foo
```

fzumstein:
got it, makes sense!

fzumstein:
I replaced this with #2656 as I couldn't merge the xlam directly and wanted to fix up the comments

PR #2668, opened 2026-01-21
Prevent type hints from breaking `Accessor` lookups.

### LINKED ISSUE #2666: Type hint analysis breaks previously working subs since v0.32.0
There are quite a few things that broke for me relating to type hints when upgrading to any version `>0.32.0`, where type hint analysis was introduced: https://docs.xlwings.org/en/stable/whatsnew.html#v0-32-0-aug-13-2024

## Example 1: Sub with return type `None`

This works:

```
@xw.sub
def foo():
    do_something()
```

But this causes an error:

```
@xw.sub
def foo() -> None:
    do_something()
```

The error happens in `def write(value, rng, options, engine_name=None):` in `xlwings/conversion/__init__.py`. This gets called with `options` like `{"convert":NoneType}`, which causes the `accessor` to be `NoneType`, which then fails when trying to do `.router` on it.

## Example 2: Sub with argument type `Iterable[str]`

This works:

```
@xw.sub
def foo(asdf):
    do_something()
```

But this causes an error:

```
@xw.sub
def foo(asdf: Iterable[str]):
    do_something()
```

When passing a VBA `Collection` of strings from VBA to my custom sub, I always have it hinted as `Iterable[str]`, since I don't care about the details of the `Collection` object, and only care about the fact that it is iterable. However, using `Iterable` as an argument type hint causes an error in `def read(rng, value, options, engine_name=None):` in `xlwings/conversion/__init__.py`, similar to above, where the `accessor` becomes the `Iterable` type itself instead of an `Accessor` class.

## Suggestion to solve this and other similar examples

When `accessors.get(...)` does not return a subclass of `Accessor`, use the default accessor like in the days before `v0.32.0`.

rudolfbyker:
Closes #2666 

Replaces #2667

### DIFF
--- a/tests/test_get_accessor.py
+++ b/tests/test_get_accessor.py
@@ -1,0 +1,119 @@
+from xlwings.conversion import (
+    _get_accessor,
+    ValueAccessor,
+    RawValueAccessor,
+    RangeAccessor,
+)
+
+
+def test_none_default() -> None:
+    """
+    When `convert` is `None`, the default Accessor class is returned.
+    """
+    assert ValueAccessor == _get_accessor(
+        convert=None,
+        default=ValueAccessor,
+        registered={
+            "raw": RawValueAccessor,
+            "range": RangeAccessor,
+        },
+    )
+
+
+def test_registered_name() -> None:
+    """
+    `convert` may be a registered name that maps to an Accessor class.
+    """
+    assert ValueAccessor == _get_accessor(
+        convert="value",
+        default=RawValueAccessor,
+        registered={
+            "range": RangeAccessor,
+            "value": ValueAccessor,
+        },
+    )
+
+
+def test_not_registered() -> None:
+    """
+    When `convert` is an unregistered name, the default Accessor class is returned.
+    """
+    assert ValueAccessor == _get_accessor(
+        convert="foo",
+        default=ValueAccessor,
+        registered={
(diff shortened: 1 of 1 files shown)


xlwings/conversion/__init__.py
```diff
@@ -78,9 +81,54 @@
...
+        convert:
+            The user-provided `convert` range option.
+            This may be one of:
+                - A registered name that maps to an Accessor class.
+                - A registered type that maps to an Accessor class.
+                - An accessor class (which will be returned directly).
+                - None, in which case the default accessor is used.
+        default:
+            The default Accessor class to use if `convert` is None or not found.
+        registered:
```
rudolfbyker: @fzumstein I deliberately refrained from using the global `accessors` dictionary here, because globals make writing independent unit tests hard. So I pass it in to the helper function instead.

xlwings/conversion/framework.py
```diff
@@ -93,8 +97,8 @@ def writer(cls, options):
 
     @classmethod
     def register(cls, *types):
-        for type in types:
-            accessors[type] = cls
+        for type_ in types:
+            accessors[type_] = cls
```
rudolfbyker: Technically unrelated to this PR, but I think it's a good rename to sneak in here.

